Issue #12, opened 2019-04-04
Include FISTA restart in pysap-mri


zaccharieramzi:
Since the restarting versions of FISTA have been included in `modopt`, we could allow the user to use them in [`sparse_rec_fista`](https://github.com/CEA-COSMIC/pysap-mri/blob/master/mri/numerics/reconstruct.py#L36) in `pysap-mri`.

chaithyagr:
I am assuming this can be closed as fixes at TOT
Please reopen if I am wrong, @zaccharieramzi

PR #14, opened 2019-04-17
added possibility to use restarting strategies

zaccharieramzi:
This solves #12 . 
I also took advantage of it to solve #5 since I was messing with `sparse_rec_fista`.

### DIFF
--- a/mri/numerics/reconstruct.py
+++ b/mri/numerics/reconstruct.py
@@ -99,14 +99,13 @@ def sparse_rec_fista(gradient_op, linear_op, prox_op, cost_op,
         print(" - alpha variable shape: ", alpha.shape)
         print("-" * 40)
 
-    # Define the proximity dual operator
-    weights = copy.deepcopy(alpha)
-    weights[...] = mu
-    prox_op.weights = weights
 
     beta_param = gradient_op.inv_spec_rad
     if lambda_update_params.get("restart_strategy") == "greedy":
         lambda_update_params["min_beta"] = gradient_op.inv_spec_rad
+        # this value is the recommended one by J. Liang in his article
+        # when introducing greedy FISTA.
+        # ref: https://arxiv.org/pdf/1807.04005.pdf
         beta_param *= 1.3
 
     # Define the optimizer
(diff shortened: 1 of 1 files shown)


mri/numerics/reconstruct.py
```diff
@@ -96,6 +99,16 @@ def sparse_rec_fista(gradient_op, linear_op, prox_op, cost_op,
         print(" - alpha variable shape: ", alpha.shape)
         print("-" * 40)
 
+    # Define the proximity dual operator
+    weights = copy.deepcopy(alpha)
+    weights[...] = mu
```
LElgueddari: The mu variable is not defined.
Please remove it

mri/numerics/reconstruct.py
```diff
@@ -96,6 +99,16 @@ def sparse_rec_fista(gradient_op, linear_op, prox_op, cost_op,
...
+    prox_op.weights = weights
+
+    beta_param = gradient_op.inv_spec_rad
+    if lambda_update_params.get("restart_strategy") == "greedy":
+        lambda_update_params["min_beta"] = gradient_op.inv_spec_rad
+        beta_param *= 1.3
```
LElgueddari: Where does this value come from, could you comment please?

mri/numerics/reconstruct.py
chaithyagr: Yes, I think thats why the Travis-build is failing. 
Also, if your updates affect the tests, can you please update them and ensure they are sane?

zaccharieramzi: Actually it's just because this part of the code is not wanted anymore, right?
We don't set the threshold within that method but actually, have it set before in the operator, right?

mri/numerics/reconstruct.py
zaccharieramzi: sure, tell me if it's enough

mri/numerics/reconstruct.py
LElgueddari: Exactly, if you really need the weights take them from the proximity operator

mri/numerics/reconstruct.py
LElgueddari: Perfect, Thanks

Issue #11, opened 2019-04-04
Include POGM' in pysap-mri


zaccharieramzi:
Since POGM' [has been included in `modopt`](https://github.com/CEA-COSMIC/ModOpt/pull/41) we could implement the corresponding sparse reconstruction function in `pysap-mri`.

zaccharieramzi:
Closed by #18

PR #18, opened 2019-04-17
Sparse rec pogm

zaccharieramzi:
Added the possibility to run the sparse reconstruction using POGM' (interesting when doing benchmarks). This solves #11 .


philouc:
Did you add a specific example or test case?

zaccharieramzi:
No I didn't.

I guess it would be a nice thing to have unit tests before even having examples for this repo.

Issue #20, opened 2019-05-03
[Bug]: Error when running sparse_rec_pogm with metrics sets to None


LElgueddari:
So I was trying to run the `sparse_rec_pogm` with metrics equal to `None` and I got the following issue
```
AttributeError: 'NoneType' object has no attribute 'items'
```
While when I run the `sparse_rec_fista` or `sparse_rec_fista_conda` everything works perfectly fine. This is coming from 
https://github.com/CEA-COSMIC/pysap-mri/blob/0fbb3f8015a27f0975f492659c02ba402c09f84e/mri/numerics/reconstruct.py#L443

where it should be set as `metric or {}` like it is in the other fista and condatvu cases

zaccharieramzi:
I think it's more in [that line](https://github.com/CEA-COSMIC/pysap-mri/blob/0fbb3f8015a27f0975f492659c02ba402c09f84e/mri/numerics/reconstruct.py#L370), where the default should not be `None` but `{}`.
Do you want me to make a PR fixing this ?

LElgueddari:
Sure and I'm also changing it in `sparse_rec_fista` and `sparse_rec_condatvu` to make it consistent

zaccharieramzi:
Well if you are changing it in those 2 you might as well open a PR for the 3. This will make things easier to review.

PR #21, opened 2019-05-03
Changing default value of `metrics`

LElgueddari:
Replace the `None`  default value on `metrics` with `{}`.
This should solve #20


zaccharieramzi:
@LElgueddari can we merge this?

Issue #19, opened 2019-05-02
Weights of the `prox_op` operator should not be set by the `sparse_rec*`


LElgueddari:
In `mri/numerics/reconstruct.py` the weights are set by the `sparse_rec*`, this is perfectly fine for basic norms such as the `ell_1` norm. However, for more sophisticated norm such as OSCAR norm, the weights are a little bit more complicated, hence they should be set by the proximity operator at the instanciation and not by the `sparse_rec*` functions

zaccharieramzi:
Totally agree.

I think the best place for the weights to be set is in [`generate_operators`](https://github.com/CEA-COSMIC/pysap-mri/blob/master/mri/reconstruct/utils.py#L108).

PR #22, opened 2019-05-03
Removing setting of the mu operator in sparse_rec functions

LElgueddari:
This improves the generalisation of the sparse_rec functions for more proximity operator methods.
This solves issues #19

### DIFF
--- a/mri/reconstruct/utils.py
+++ b/mri/reconstruct/utils.py
@@ -120,7 +120,7 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
         the wavelet name to be used during the decomposition.
     samples: np.ndarray
         the mask samples in the Fourier domain.
-    mu: float, (defaul=1e-06)
+    mu: float, (defaul=1e-06) or np.ndarray
         Regularization hyper-parameter should be positif
     nb_scales: int, default 4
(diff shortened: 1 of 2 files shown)


mri/reconstruct/utils.py
```diff
@@ -188,7 +192,7 @@ def generate_operators(data, wavelet_name, samples, nb_scales=4,
             fourier_op=fourier_op)
 
     # Define the proximity dual/primal operator
-    prox_op = SparseThreshold(linear_op, None, thresh_type="soft")
+    prox_op = SparseThreshold(linear_op, mu, thresh_type="soft")
```
zaccharieramzi: It has to be at the right shape, so you need a vector not a float.

LElgueddari: I added the case where the hyper-parameter is an array

LElgueddari: I don't really understand why the my should be an array if the `SparseThreshold` works fine with a float, it is more efficient in terms of memory to have a float than an array

zaccharieramzi: no you are right it can be a float. I was just confuse because we had always been setting it with an array, but actually I checked and it can indeed by just a float, my bad.

zaccharieramzi:
@LElgueddari can we merge this?

PR #25, opened 2019-09-09
Added Support for 3D Reconstruction

chaithyagr:
This branch generalizes the flow for 3D reconstruction.
I have added an example python script too.
I have seen that sometimes the cost increases if we don't have sufficient iterations in the calculation of Lipschitz Constant. Any ideas on how to fix this?

PR #26, opened 2019-09-09
Added support for Calibrationless

chaithyagr:
This holds the first set of codes that work fine syntactically.
I have seen the cost function decrease for now, but need to work in-depth on the actual output


chaithyagr:
We have an ongoing PR #25 , and we plan to get at this once we have that in.. Closing this without merging.

PR #29, opened 2019-09-11
Fix Coverage issues

chaithyagr:
This PR fixes the coverage issues earlier which did not count right.
We are now at 53% of coverage now.
@LElgueddari , we may need more tests, but that will be a separate PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ install:
       fi
 
 script:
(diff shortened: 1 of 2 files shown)


chaithyagr:
I tried that method too, somehow it didn't catch any tests. 
I'll check it again locally.

chaithyagr:
There were many minor issues while using nosetests.
I have moved to use pytest and we work very similar to modopt now. I have added the setup.cfg now.

Issue #33, opened 2019-09-18
Implement stacked version of fourier transform for 3D data having stack of same k-space data


chaithyagr:
The NFFT-3D implementation takes too long for running at high N's
The idea is that if the input data is a stack of data at same kspace locations at different z-levels, we could exploit this to obtain the 3D NFFT faster by implementing a 2D NFFT followed by a final 1D - FFT.

Currently, I have base codes up and working for this, the seen speedup is as follows:

For N 64 Speedup = 0.47019157789921967
For N 128 Speedup = 0.7234951284094772
**For N 256 Speedup = 3.391414416287289
For N 338 Speedup = 3.5204348276982196
For N 448 Speedup = 4.978218234526903**

Clearly, there is a lot of merit in getting these codes to mainline for specific cases as mentioned above.
I will document my code now and get a PR for this.

chaithyagr:
Also, for this feature, I have added check codes to ensure that the incoming samples are in the order as needed. Further, I have added 2 tests for this, namely for checking the adjoint property and to check that the output is close to that of 3D NFFT.

chaithyagr:
This has been implemented and seems to be a stale issue. Closing

PR #34, opened 2019-09-18
Stacked 3D implementation of NFFT3D

chaithyagr:
This addresses the feature discussed in #33 
Also some code cleanups for adjoint tests


mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+    shape: tuple of int
+        shape of the image (necessarly a square/cubic matrix).
+    fft_type: 'NFFT' | 'NUFFT' default 'NFFT'
+        What FFT version to be used for 2D non uniform FFT
+    platform: string, 'cpu', 'multi-cpu' or 'gpu' default 'gpu'
+        string indicating which hardware platform will be used to compute the
+        NUFFT. works only if fft_type=='NUFFT'
+    """
+
+    def __init__(self, samples, shape, fft_type='NFFT', platform='gpu'):
```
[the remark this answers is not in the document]
chaithyagr: But, if we do that, it would mean that everytime we initialize a NUFFT, we need to pass the argument `platform='gpu'`, which seems inefficient to me. 
What we could possibly do is that keep platform as GPU by default for NUFFT and get rid of this parameter altogether, how does that sound?

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+            the mask samples in the Fourier domain.
+        shape: tuple of int
+            shape of the image (necessarly a square/cubic matrix).
+        fft_type: 'NFFT' | 'NUFFT' default 'NFFT'
+            What FFT version to be used for 2D non uniform FFT
+        platform: string, 'cpu', 'multi-cpu' or 'gpu' default 'gpu'
+            string indicating which hardware platform will be used to
+            compute the NUFFT. works only if fft_type=='NUFFT'
+        """
+        self.num_slices = shape[2]
```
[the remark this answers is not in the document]
chaithyagr: I think this is covered in my above comment, we could just pass kwargs to NUFFT object which is given by user

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+    """
+
+    def __init__(self, samples, shape, fft_type='NFFT', platform='gpu'):
+        """ Init function for Stacked3D class
+        Parameters
+        ----------
+        samples: np.ndarray
+            the mask samples in the Fourier domain.
+        shape: tuple of int
+            shape of the image (necessarly a square/cubic matrix).
```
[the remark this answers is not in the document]
chaithyagr: I think Ill correct it to `stack in 3D` . Does that sound better?

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+        string indicating which hardware platform will be used to compute the
+        NUFFT. works only if fft_type=='NUFFT'
+    """
+
+    def __init__(self, samples, shape, fft_type='NFFT', platform='gpu'):
+        """ Init function for Stacked3D class
+        Parameters
+        ----------
+        samples: np.ndarray
+            the mask samples in the Fourier domain.
```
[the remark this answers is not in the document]
chaithyagr: Yes, thats better

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+        shape: tuple of int
+            shape of the image (necessarly a square/cubic matrix).
+        fft_type: 'NFFT' | 'NUFFT' default 'NFFT'
+            What FFT version to be used for 2D non uniform FFT
```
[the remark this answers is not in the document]
chaithyagr: Hmm, makes sense

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+        platform: string, 'cpu', 'multi-cpu' or 'gpu' default 'gpu'
+            string indicating which hardware platform will be used to
+            compute the NUFFT. works only if fft_type=='NUFFT'
+        """
+        self.num_slices = shape[2]
+        self.shape = shape
```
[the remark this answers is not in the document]
chaithyagr: Ahhm that was for debug, will remove

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
         """
         self.plan.f = x
         return np.copy(self.plan.adjoint()) / np.sqrt(self.plan.M)
+
+
+class Stacked3D:
```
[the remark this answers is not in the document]
chaithyagr: Yes, will get this..

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+        samples = samples[self.sort_pos]
+        plane_samples, self.z_samples = self.get_stacks(samples)
+        if fft_type == 'NUFFT':
+            self.FT = NUFFT(samples=plane_samples, shape=shape[0:2],
+                            platform=platform)
+        elif fft_type == 'NFFT':
+            self.FT = NFFT(samples=plane_samples, shape=shape[0:2])
+
+    def get_stacks(self, samples):
+        """ Function that converts an incoming 3D kspace samples
```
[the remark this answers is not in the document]
chaithyagr: Makes sense, just that checking for stacks involves actually getting them all and sending one of them. But , I will make it a different function

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+        self.num_slices = shape[2]
+        self.shape = shape
+        # Sort the incoming data based on Z, Y then X coordinates
+        # This is done for easier stacking
+        self.sort_pos = np.lexsort(tuple(samples[:, i]
+                                         for i in np.arange(3)))
+        samples = samples[self.sort_pos]
+        plane_samples, self.z_samples = self.get_stacks(samples)
+        if fft_type == 'NUFFT':
+            self.FT = NUFFT(samples=plane_samples, shape=shape[0:2],
```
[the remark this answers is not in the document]
chaithyagr: Covered above

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
         """
         self.plan.f = x
         return np.copy(self.plan.adjoint()) / np.sqrt(self.plan.M)
+
+
+class Stacked3D:
+    """"  3-D non uniform Fast Fourrier Transform class,
+        fast implementation for Stacked samples
```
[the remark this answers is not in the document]
chaithyagr: Hmm, isnt there a lot , even with description of what is stack ? 
Could you please help me here?

mri/reconstruct/fourier.py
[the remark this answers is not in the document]
chaithyagr: We may have to merge #30  to cover this for now...

mri/reconstruct/fourier.py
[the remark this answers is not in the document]
LElgueddari: I am not sure about putting the GPU by default in the NUFFT because of two things:
First, the GPU version of the NUFFT does not come with the installation of the plugin.
Second, no travis tests have been made with the GPU version.
For those two reasons, I think we should keep the default setting to `'cpu'`, hence someone who wanted to use the GPU needs to specify it.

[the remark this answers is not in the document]
chaithyagr: Wait, I think here was my idea:
1) Use NFFT for CPU, 
2) Use NUFFT for GPU or Multiple-CPU

I dont understand why we would have to use NUFFT with CPU. But you do make a sound point that isnt tested, what is the best way to go with here?

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
         """
         self.plan.f = x
         return np.copy(self.plan.adjoint()) / np.sqrt(self.plan.M)
+
+
+class Stacked3D:
+    """"  3-D non uniform Fast Fourrier Transform class,
```
philouc: check typos: Fourrier -> Fourier

chaithyagr: Ahh, thank you @philouc . I will get this fixed in the next commit.

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+    def op(self, data):
+        """ This method calculates Fourier transform.
+        Parameters
+        ----------
+        img: np.ndarray
+            input image as array.
+
+        Returns
+        -------
+        result: np.ndarray
```
zaccharieramzi: I wasn't paying attention in the last PR probably, but expliciting a name for the return value isn't a good practice imho, especially when it doesn't match the name in code (happens in other places in this PR).

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+            Forward 3D Fourier transform of the image.
+        """
+        first_fft = np.fft.fftshift(np.fft.fft(np.fft.ifftshift(data, axes=2),
+                                               n=self.acq_num_slices,
+                                               norm="ortho"),
+                                    axes=2)
+        final = np.asarray([self.FT.op(first_fft[:, :, slice])
```
zaccharieramzi: `final` may not be the best variable name.

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+                            for slice in np.arange(self.acq_num_slices)])
+        final = np.reshape(final, self.acq_num_slices * self.first_stack_len)
+        # Unsort the Coefficients and send
```
zaccharieramzi: send ?

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+        if fft_type == 'NUFFT':
+            self.FT = NUFFT(samples=plane_samples, shape=shape[0:2],
+                            platform=platform)
+        elif fft_type == 'NFFT':
+            self.FT = NFFT(samples=plane_samples, shape=shape[0:2])
+
+    def get_stacks(self, samples):
+        """ Function that converts an incoming 3D kspace samples
+            and converts to stacks of 2D. This function also checks for
+            any issues incoming k-space pattern and if the stack property
```
zaccharieramzi: typo I think: 'any issues of the incoming k-space'

mri/reconstruct/fourier.py
```diff
@@ -208,3 +208,139 @@ def adj_op(self, x):
...
+            is not satisfied.
+            Stack Property:
+                The k-space locations originate from a stack of 2D samples
+        Parameters
+        ----------
+        samples: np.ndarray
```
zaccharieramzi: I think they are not samples or mask samples right? they are sample locations.
I think this is in-line with what @LElgueddari  said in this [comment](https://github.com/CEA-COSMIC/pysap-mri/pull/34/files#r326042782), but it happens throughout the code (both in docs and in variable names)

mri/reconstruct/fourier.py
chaithyagr: Hmm, I was not sure what to give, Ill name is result? I am a bit clueless on this..

mri/reconstruct/fourier.py
LElgueddari: Why isn't that a good practice?

chaithyagr: Hmm, I don't exactly follow your point or don't agree with it. 
Are you saying that in documentation we should not give the name of return variable? 

In my opinion, we need to give such names and also tell what the variable is. The name is very important especially when the function returns multiple variables

mri/reconstruct/fourier.py
chaithyagr: `return` sounds better?

mri/reconstruct/fourier.py
chaithyagr: yes, sorry

mri/reconstruct/fourier.py
chaithyagr: This seems to be like a legacy which I followed. The idea is that from all the Fourier operators viewpoint, samples mean where the samples come from. 
AT a point this will become a philosophical conversation :P , but we can decide what we want to name it and I could fix it at all places in this same PR with string replace.
@LElgueddari @zaccharieramzi , to comment

mri/reconstruct/fourier.py
zaccharieramzi: I think you just don't need it

mri/reconstruct/fourier.py
zaccharieramzi: well what is it?

chaithyagr: It is the **final** K-space data which is stacked. Hence the name **final** :P

mri/reconstruct/fourier.py
zaccharieramzi: It's not necessarily a bad practice, but you just don't need it, you need the names of the arguments because they can be used to call the function, you don't need to know the name of the return variable which is internal to the function. In short it's not needed, so we shouldn't have (but I am not strongly against it).

When the function returns multiple variables, it might be better for clarity's sake. Even still, the name should match that of the returned variable.

mri/reconstruct/fourier.py
zaccharieramzi: why not `stacked_kspace` then?

mri/reconstruct/fourier.py
zaccharieramzi: can you point me to where you saw that "samples mean where the samples come from".
I think it's not a philosophical conversation. To me the difference between samples and sample locations is pretty clear, and confusing both is often dangerous or misleading.

chaithyagr: Will open a newe issue for this, I will not cover this in this PR

zaccharieramzi: why?

chaithyagr: The PR is very bulky already. I feel it makes sense to handle this as a separate issue

zaccharieramzi: I don't understand. To me you are introducing a misleading variable name, you need to correct it. It doesn't change the bulk of this PR, just corrects it.

chaithyagr: Well, I can fix in this Class surely, but across to all other classes too? That is what I am saying will not be done in this PR

zaccharieramzi: yes not necessarily in other classes just start here. What are these other classes?

chaithyagr: NFFT / NUFFT all use samples as a parameter

mri/reconstruct/fourier.py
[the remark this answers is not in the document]
chaithyagr: @LElgueddari ping? What more shall we add?

mri/reconstruct/fourier.py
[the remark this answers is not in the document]
chaithyagr: This is void now, given we use NonCartesian FFT

mri/reconstruct/fourier.py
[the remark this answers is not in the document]
chaithyagr: I have moved this to utils as this can be used for even Stack based sensitivity extraction

mri/reconstruct/fourier.py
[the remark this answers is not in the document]
LElgueddari: Yes just saying that fo each stack the samples should be the same, cause you might have some stack version with sampling distribution different from one stack to another

LElgueddari:
Do you think you could also add an example for the stack 3D thing?

mri/reconstruct/fourier.py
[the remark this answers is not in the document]
chaithyagr: > Do you think you could also add an example for the stack 3D thing?

Let me see what I can do. I'll mostly use Orange and write a Reconstruction with a Stacked mask

chaithyagr:
> Do you think you could also add an example for the stack 3D thing?

Yes, I plan to add one, for stack based reconstruction. (Non Cartesian, using radial)

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -93,6 +94,43 @@ def gridded_inverse_fourier_transform_nd(kspace_loc,
...
+                                            stack_len, kspace_data, grid,
+                                            method):
+    """
+    This function calculates the gridded Inverse fourier transform
+    from Interpolated non-Cartesian data into a cartesian grid. However,
+    the IFFT is done similar to Stacked FOurier transform.
+
+    Parameters
+    ----------
+    kspace_loc: np.ndarray
```
zaccharieramzi: the doc doesn't correspond to the function parameters (I think `kspace_loc` is beter than `plane_samples`, it gives a better indication of what this variable is, i.e. positions)

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -93,6 +94,43 @@ def gridded_inverse_fourier_transform_nd(kspace_loc,
...
+        documentation
+    grid: np.ndarray
+        The Gridded matrix for which you want to calculate k_space Smaps
+    Returns
+    -------
+    np.ndarray
+        The gridded inverse fourier transform of given kspace data
+    """
+    gridded_kspace = []
+    for i in np.arange(len(z_samples)):
```
zaccharieramzi: nitpicking but why `np.arange` instead of `range`?

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -93,6 +94,43 @@ def gridded_inverse_fourier_transform_nd(kspace_loc,
...
+        gridded_kspace.append(
+            griddata(plane_samples,
+                     kspace_data[i*stack_len:(i+1)*stack_len],
+                     grid,
+                     method=method,
+                     fill_value=0))
+    gridded_kspace = np.moveaxis(np.asarray(gridded_kspace), 0, 2)
```
zaccharieramzi: Could you add a comment on all these changes of axis (`moveaxis`, `swapaxes`)?

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,114 @@ def adj_op(self, coeffs):
...
+        Parameters
+        ----------
+        kspace_loc: np.ndarray
+            the position of the samples in the k-space
+        shape: tuple of int
+            shape of the image stack in 3D. (N x N x Nz)
+        implementation: string, 'cpu', 'cuda' or 'opencl' default 'cpu'
+            string indicating which implemenmtation of Noncartesian FFT
+            must be carried out. Please refer to Documentation of
+            NoncartesianFFT
```
zaccharieramzi: You are missing `n_coils` docs

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,114 @@ def adj_op(self, coeffs):
...
+        """
+        self.num_slices = shape[2]
+        (plane_samples, self.z_samples, self.stack_len,
+         self.acq_num_slices, self.sort_pos) = get_stacks_fourier(kspace_loc)
+        self.FT = NonCartesianFFT(samples=plane_samples, shape=shape[0:2],
+                                  implementation=implementation)
+        self.nb_coils = n_coils
```
zaccharieramzi: Why not the same name?

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,114 @@ def adj_op(self, coeffs):
...
+        # Unsort the Coefficients
+        inv_idx = np.zeros_like(self.sort_pos)
+        inv_idx[self.sort_pos] = np.arange(len(self.sort_pos))
+        return stacked_kspace[inv_idx]
+
+    def op(self, data):
+        """ This method calculates Fourier transform.
+        Parameters
+        ----------
+        img: np.ndarray
```
zaccharieramzi: the parameter name doesn't match

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,114 @@ def adj_op(self, coeffs):
...
+        #                               axis=0, bounds_error=False,
+        #                               fill_value=0)
+        # first_fft = interpolate_kspace(np.linspace(-0.5, 0.5,
+        #                                            self.num_slices,
+        #                                            endpoint=False))
+        final = np.fft.ifftshift(np.fft.ifft(
+            np.asarray(np.fft.fftshift(first_fft, axes=0)),
+            axis=0, n=self.num_slices, norm="ortho"),
+            axes=0)
+        return final.T
```
zaccharieramzi: can you add a comment on why we transpose?

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,114 @@ def adj_op(self, coeffs):
...
+
+    def adj_op(self, coeff):
+        """ This method calculates inverse masked non-uniform Fourier
+        transform of a 1-D coefficients array.
+        Parameters
+        ----------
+        x: np.ndarray
```
zaccharieramzi: watch out for all the parameters/docs mismatches (this one is the last one I highlight but may be more down the road)

mri/reconstruct/utils.py
```diff
@@ -233,3 +233,46 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
...
+    ----------
+    samples: np.ndarray
+        the mask samples in the 3D Fourier domain.
+
+    Returns
+    ----------
+    plane_samples: np.ndarray
+        A 2D array of samples which when stacked gives the 3D samples
+    z_samples: np.ndarray
+        A 1D array of z-sample locations
```
zaccharieramzi: can you call it `z_samples_loc` then?

mri/parallel_mri/extract_sensitivity_maps.py
chaithyagr: I dont have any prefrence, its kinda random, will stick to range

mri/parallel_mri/extract_sensitivity_maps.py
chaithyagr: Yes on it

mri/reconstruct/fourier.py
chaithyagr: Oops. Will add

mri/reconstruct/utils.py
chaithyagr: Yes. Ill use `z_sample_loc` and `kspace_plane_loc`

mri/reconstruct/fourier.py
chaithyagr: Legacy, Will keep it same here and will fix in other cases going forward

chaithyagr:
@LElgueddari. I have a basic example that I formulated, I am having some issues which I will work through, but irrespective, we can get this feature in if it takess too long, especially given that we dont have basic 3D examples yet. For sanity, I have tests in place

mri/reconstruct/utils.py
```diff
@@ -218,18 +228,64 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
...
     # TODO need to have multiple cost functions with a parameter
     cost_op = GenericCost(
         gradient_op=gradient_op,
         prox_op=prox_op,
         verbose=verbose)
     return gradient_op, linear_op, prox_op, cost_op
+
+
+def get_stacks_fourier(kspace_loc):
+    """ Function that converts an incoming 3D kspace samples
```
LElgueddari: I rather change this sentence for something like:
```suggestion
    """ Function that splits an incoming 3D stacked k-space samples 
    into a 2D non-Cartesian plane and the vector containing the z k-space values 
    of all the plane 
```

mri/test/test_fourier_adjoint.py
```diff
@@ -155,6 +157,70 @@ def test_NFFT_3D(self):
                 np.testing.assert_allclose(x_d, x_ad, rtol=1e-10)
         print(" NFFT in 3D adjoint test passes")
 
+    def test_adjoint_stack_3D(self):
```
LElgueddari: You should also consider testing the error. 
```    
def test_stack_3D_error(self):
    import numpy.testing as npt
    npt.assert_raises(ValueError, get_stacks_fourier, np.random.randn(12, 3)
```
Or something like that

mri/test/test_fourier_adjoint.py
```diff
@@ -155,6 +157,70 @@ def test_NFFT_3D(self):
...
+    def test_similarity_stack_3D(self):
+        """Test the similarity of stacked implementation of Fourier transform
+        to that of NFFT
+        """
+        for channel in self.num_channels:
+            print("Testing with num_channels=" + str(channel))
+            for N in [64, 128]:
+                # Nz is the number of slices, this would check both N=Nz
+                # and N!=Nz
+                Nz = 64
```
LElgueddari: Just for computation efficiency could you put a lower value of `Nz` like 2 or 4?

chaithyagr: Well I need NFFT to also run, and thus I will keep it at Nz=16 (NFFT is unstable at lower Nz). I will test with N=16 and N=32

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,125 @@ def adj_op(self, coeffs):
             inverse discrete Fourier transform of the input coefficients.
         """
         return self.implementation.adj_op(coeffs)
+
+
+class Stacked3D(FourierBase):
```
zaccharieramzi: Probably should be name `Stacked3DNFFT`

mri/reconstruct/gradient.py
```diff
@@ -20,26 +20,26 @@
 
 class GradAnalysis2(GradBasic, PowerMethod):
 
-    def __init__(self, data, fourier_op):
+    def __init__(self, data, fourier_op, max_iter):
```
zaccharieramzi: I think `max_iter` in this context is a bit misleading. Ppl might think the gradient is computed iteratively when it's not. You can either rename the variable or add a doc. I think the best way is to rename the variable to something like `max_iter_spec_rad`.

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -93,6 +94,49 @@ def gridded_inverse_fourier_transform_nd(kspace_loc,
...
+    from Interpolated non-Cartesian data into a cartesian grid. However,
+    the IFFT is done similar to Stacked FOurier transform.
+
+    Parameters
+    ----------
+    kspace_plane_loc: np.ndarray
+        The N-D k_space locations of size [M, N]. These hold locations only
+        in plane, extracted using get_stacks_fourier function
+    z_samples: np.ndarray
+        This holds the z-sample locations for stacks. Again, extracted using
```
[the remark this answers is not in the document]
chaithyagr: I thought I had updated this :( , mostly missed this..

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,125 @@ def adj_op(self, coeffs):
...
+    """
+
+    def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
+        """ Init function for Stacked3D class
+        Parameters
+        ----------
+        kspace_loc: np.ndarray
+            the position of the samples in the k-space
+        shape: tuple of int
+            shape of the image stack in 3D. (N x N x Nz)
```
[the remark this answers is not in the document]
chaithyagr: Ack

mri/reconstruct/fourier.py
chaithyagr: Cool

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,125 @@ def adj_op(self, coeffs):
...
+        (kspace_plane_loc, self.z_sample_loc,
+         self.sort_pos) = get_stacks_fourier(kspace_loc)
+        self.acq_num_slices = len(self.z_sample_loc)
+        self.stack_len = len(kspace_plane_loc)
+        self.FT = NonCartesianFFT(samples=kspace_plane_loc, shape=shape[0:2],
+                                  implementation=implementation)
+        self.n_coils = n_coils
+
+    def _op(self, data):
+        first_fft = np.fft.fftshift(np.fft.fft(np.fft.ifftshift(data, axes=2),
```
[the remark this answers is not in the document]
chaithyagr: Yes will change

mri/reconstruct/fourier.py
```diff
@@ -521,3 +521,125 @@ def adj_op(self, coeffs):
...
+            Number of coils used to acquire the signal in case of multiarray
+            receiver coils acquisition
+        """
+        self.num_slices = shape[2]
+        self.shape = shape
+        (kspace_plane_loc, self.z_sample_loc,
+         self.sort_pos) = get_stacks_fourier(kspace_loc)
+        self.acq_num_slices = len(self.z_sample_loc)
+        self.stack_len = len(kspace_plane_loc)
+        self.FT = NonCartesianFFT(samples=kspace_plane_loc, shape=shape[0:2],
```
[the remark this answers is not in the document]
chaithyagr: cool

mri/reconstruct/gradient.py
chaithyagr: Will change to `max_iter_spec_rad`

chaithyagr:
> General question (probs should have asked earlier but I just realized). Why do we only support Stacked non-uniform 3D Fourier transform?

I dont think I understand this @zaccharieramzi . The reason is speedups and also that it is very memory efficient

zaccharieramzi:
My question was why don't we also support Stacked 3D (regular) Fourier Transform?

LElgueddari:
So basically the idea of stack 3d is to take advantage of the Cartesian distribution in the 3D dimension by using FFT instead of NFFT in that direction.
For 3D FFT either stack or not the Fourier transform is already doing 1d FFT in each direction. Therefore stack 3D fft is just the basic fftn. Hence there is no point of doing that

philouc:
> My question was why don't we also support Stacked 3D (regular) Fourier Transform?

My answer @zaccharieramzi : because we're supporting 3D acquisitions that encode stack of stars (radial)/spirals/2D sparkling and then along the k_z direction, the sampling is regular  so FFT is much more efficent as @LElgueddari outlined

zaccharieramzi:
I understand that `fftn` allows us to perform 3D stacked FFT, I am asking why we don't include it. For example we include 2D FFT even if `fft2` already allows to do it.

chaithyagr:
@LElgueddari gave a spot on answer there.. 
Here is fftn implementation which does exactly the same over all axes:

https://github.com/numpy/numpy/blob/d9b1e32cb8ef90d6b4a47853241db2a28146a57d/numpy/fft/pocketfft.py#L624-L631

zaccharieramzi:
So there is absolutely no need for stacked cartesian @philouc ?
@chaithyagr like I said, I understand `fftn`, but we do support `fft2` so why not `fftn`?

zaccharieramzi:
@chaithyagr I see thanks, that was the answer I was looking for.

philouc:
Yes we could add support to 3D Cartesian sampling/undersampling as it is used for instance in 3D EPI for fMRI.

LElgueddari:
PR #51 already support 3D

chaithyagr:
All checks have passed! @LElgueddari can you please merge?

Issue #35, opened 2019-09-19
Adding examples


LElgueddari:
Could we add examples to the MRI plugins this would be much easier to test and integrate the new features developed? 
Also, exemples could be tested with [CircleC](https://ci.appveyor.com/login)I to make sure that they are still running perfectly when we do changes?

chaithyagr:
> Also, exemples could be tested with CircleCI to make sure that they are still running perfectly when we do changes?

How do we plan on doing this? I mean we can have examples run, but what will be the check criteria to ensure it is running right?

chaithyagr:
Listing out Examples that we could have here, please add / remove on to the list:

{Cartesian / Non Cartesian} x {2D} x {Single Channel, Multichannel}
For Multichannel: {Self Calibrating and Calibrationless}
**Number = 6**

{Non Cartesian} x {3D} x {Single Channel, Multichannel}
For Multichannel: {Self Calibrating and Calibrationless}
**Number = 3**

{2D/3D} x Dictionary Learning.
**Number = 2**

**TOTAL = 11**

chaithyagr:
As a first step, I will get a PR for following examples:
1) 2D Single Channel Cartesian with:
   * Linear Operator: PyWt Decimated Wavelet (Orthogonal, or Bi-Orthogonal Wavelet)
   * Optimization: With POGM, Condat-Vu, and Greedy FISTA

2) 2D Single Channel Non-Cartesian with:
   * Linear Operator: ISAP Undecimated Wavelet (Implement with rotated filters if necessary)
   * Optimization: same as above.


@LElgueddari , do you want to add or modify any?

zaccharieramzi:
I think one optimisation algorithm should be enough for the purpose of the paper. If we want to showcase the possibilities to gain speed using different algorithms we can just refer to my benchmark.

chaithyagr:
I think we have satisfied the purpose of this issue.
Re-purposing now to ensure we have one example for each type of reconstruction.

chaithyagr:
I think this can be closed now as we have one example of each kind.
Closing.

Issue #36, opened 2019-09-19
Code refactoring


LElgueddari:
Could we refactor the code just to have more explicit module names? 
For example, `reconstruct` module does not mean anything since the goal of the plugins is to do reconstruction.
An idea would be to split the modules according to their function i.e. `gradients`, `linear_operator` etc..

chaithyagr:
I will get an updated code structure and update here before we make any changes.
We should ideally get at this after the #35 issue is fixed so that we have something to check the stability.

chaithyagr:
@LElgueddari  and @chaithyagr  had a discussion on this and here is the set proposal that we have in mind.
@LElgueddari can you please give a thumbs up if you feel I havent missed out anything, or just point to out here.
@zaccharieramzi can you please review this and ensure we have not missed out on anything?

- mri
    - operators
        - _linear
             - dictionary
             - wavelet
        - _fourier
             - stack
             - cartesian
             - non-cartesian
                 - cpu / cuda / opencl
        - _gradient
        - _proximity

   - reconstructors
        - _base_reconstructor
            - self calibrating
                - L1 ESPIRIT based estimation
                - kspace center based estimation 
            - calibrationless (base class with common stuff, all below are individual classes as each have different parameters)
                - group LASSO
                - OWL
                - K-support Norm
                - Low Rank
                - etc..
            - single channel reconstructor

   - optimization
       - FISTA
       - CondatVu
       - POGM

   - utils
       - plotting : Holds basic 3D and 2D plotting function, to reduce dependence on pysap whose import takes too long.



We will have utils in each folder based on where it belongs to. 
Examples will remain same, we plan to add unit tests to respective folders than maintain it in a common folder. pytest will still find all the tests and run them right.


Going forward, we would want the base classes for _linear and _fourier to be moved into a different location given that many other plugins would use it.

chaithyagr:
Logistics:
We plan to freeze any PRs while we are going through the process of refactoring to ensure we dont have pile up of new codes to refactor. This PR be primarily done by me and @LElgueddari (assigned).
We will have our own branch where we start on this task and have a common PR opened.

zaccharieramzi:
What is the base_reconstructor? 
What will the optimization folder hold exactly? (The algorithms themselves are in Modopt)

chaithyagr:
_base_reconstructor holds the base class which has all the common details for reconstruction, like dimension, number of channels etc. actual reconstructors derive from this class.

[...]

PR #38, opened 2019-09-30
[SMALL] Add support for Undecimated Wavelet Transform, with right adjoint operator

chaithyagr:
This is a minor PR that will hold the undecimated Wavelet transform class for 2D, which pulls in the transform filters from Sparse2D with help of Modopt.

This is a temporary solution while we wait for general solution from https://github.com/CEA-COSMIC/pysap/issues/106 

This PR is particularly important to get the first merge for #35 with cartesian and non-cartesian reconstructions.


mri/reconstruct/linear.py
```diff
@@ -114,13 +115,162 @@ def l2norm(self, shape):
...
+        self.wavelet_id = wavelet_id
+        self.multichannel = multichannel
+        self.nb_scale = nb_scale
+        self._opt = [
+            '-t{}'.format(self.wavelet_id),
+            '-n{}'.format(self.nb_scale),
+        ]
+        self._has_run = False
+        self.coeffs_shape = None
+        self.flatten = flatten
```
zaccharieramzi: Why do you define them as object properties. I think it's fine to keep them as functions imported

mri/reconstruct/linear.py
```diff
@@ -114,13 +115,162 @@ def l2norm(self, shape):
...
+    def __init__(self, wavelet_id=24, nb_scale=4,
+                 verbose=0, multichannel=False):
+        """Init function for Undecimated wavelet transform
+
+        Parameters
+        -----------
+        wavelet_id: int, default 24 = undecimated (bi-) orthogonal transform
+            ID of wavelet being used
+        nb_scale: int, default 4
+            the number of scales in the decomposition.
```
zaccharieramzi: Don't forget to document all the args

mri/reconstruct/linear.py
```diff
@@ -114,13 +115,162 @@ def l2norm(self, shape):
...
+                    filter_convolve(data[channel].real, self.transform))
+                coefs_imag, coeffs_shape = self.flatten(
+                    filter_convolve(data[channel].imag, self.transform))
+                coeffs.append(coefs_real + 1j * coefs_imag)
+                self.coeffs_shape.append(coeffs_shape)
+            return np.asarray(coeffs)
+        else:
+            coefs_real = filter_convolve(data.real, self.transform)
+            coefs_imag = filter_convolve(data.imag, self.transform)
+            coeffs, self.coeffs_shape = self.flatten(
```
zaccharieramzi: Why would we need to flatten the coefficients?

mri/test/test_wavelet_adjoint.py
```diff
@@ -80,6 +80,23 @@ def test_Wavelet3D_PyWt(self):
             numpy.testing.assert_allclose(x_d, x_ad, rtol=1e-6)
         print(" Wavelet3 adjoint test passes")
 
+    def test_Wavelet_UD_2D(self):
+        """Test the adjoint operation for Undecimated wavelet
+        """
+        for i in range(self.max_iter):
+            print("Process Wavelet Undecimated test '{0}'...", i)
+            wavelet_op_adj = WaveletUD2(nb_scale=4)
```
zaccharieramzi: why adj?

chaithyagr:
Adding, I see that Travis Build also has failed due to missing Sparse2D installations. I will try to get this fixed here as we may need this in the future also.
Note that going with ModOpt based way to obtain MR Filters, we will face this issue, the installations and environment variables must be rightly set.

zaccharieramzi:
Yes that will have to be explained in the docs somewhere too.
The fix is something like the following:
`export PATH=$PATH:/path-to-pysap/build/temp.linux-x86_64-3.6/extern/bin/`

mri/reconstruct/linear.py
chaithyagr: Well, going forward, most of the proximity operators (especially LASSO, OWL etc, rely on this type of formatting)  I am merely ensuring consistency in the way `WaveletN` and `WaveletUD2` classes work

mri/reconstruct/linear.py
chaithyagr: Ack.

mri/reconstruct/linear.py
chaithyagr: Yes

mri/test/test_wavelet_adjoint.py
chaithyagr: Copied code from earlier test where it made sense. Fixing

mri/reconstruct/linear.py
zaccharieramzi: I guess now is not really the time to talk about it nor the place, but I think it would be nicer if it was the proximity operators job to perform the flattening.

chaithyagr: Yes, I completely agree with you. All I am saying is that we would have wider impact with it, and if we plan to do it, we need to do it everywhere (ie waveletN also). For now, do you agree that we need consistency here?

zaccharieramzi: Yes like I said it's not the time or place (that's why I marked the conversation as solved)

chaithyagr:
Adding @AGrigis to review the travis code updates. 

@AGrigis : 
Note that this line is temporary:
`    - pip install git+https://github.com/chaithyagr/ModOpt.git`

We are awaiting https://github.com/CEA-COSMIC/ModOpt/pull/64 to be merged to mainline so that we can directly point to CEA_COSMIC master branch.

mri/reconstruct/linear.py
```diff
@@ -16,9 +16,11 @@
 import pysap
 from pysap.base.utils import flatten
 from pysap.base.utils import unflatten
+from modopt.signal.wavelet import get_mr_filters, filter_convolve
 
 # Third party import
-import numpy
+import numpy as np
+from joblib import Parallel, delayed
```
zaccharieramzi: Next time for the imports don't forget to have them in alphabetical order.

mri/reconstruct/linear.py
```diff
@@ -114,13 +116,185 @@ def l2norm(self, shape):
...
+        wavelet_id: int, default 24 = undecimated (bi-) orthogonal transform
+            ID of wavelet being used
+        nb_scale: int, default 4
+            the number of scales in the decomposition.
+        multichannel: bool, default False
+            Boolean value to indicate if the incoming data is from
+            multiple-channels
+        n_cpu: int, default 0
+            Number of CPUs to run on. Only applicable if multichannel=True.
+        Private Variables:
```
zaccharieramzi: Why list those? And why only one?

mri/reconstruct/linear.py
```diff
@@ -114,13 +116,185 @@ def l2norm(self, shape):
...
         # Compute the L2 norm
-        return numpy.linalg.norm(data)
+        return np.linalg.norm(data)
+
+
+class WaveletUD2(object):
+    """The wavelet undecimated operator using pysap wrapper.
+    """
+    def __init__(self, wavelet_id=24, nb_scale=4,
+                 multichannel=False, n_cpu=1, verbose=0):
```
zaccharieramzi: add docs also for verbose, explaining what the different levels are.

mri/reconstruct/linear.py
```diff
@@ -114,13 +116,185 @@ def l2norm(self, shape):
...
+        coeffs: ndarray
+            the wavelet coefficients.
+        """
+        if not self._has_run:
+            if self.multichannel:
+                self._get_filters(list(data.shape)[1:])
+            else:
+                self._get_filters(data.shape)
+        if self.multichannel:
+            coeffs, self.coeffs_shape = zip(*Parallel(n_jobs=self.n_cpu)(
```
zaccharieramzi: Have you tried to see which backend was the fastest?
It might be interesting to leave that option to the user, given that it's not trivial determining which will be depending on the size of the data and the number of channels.
Look at this SO question for more on this: https://stackoverflow.com/questions/55956447/how-to-parallelize-scipy-fftconvolve-using-joblib

mri/test/test_wavelet_adjoint.py
```diff
@@ -80,6 +85,40 @@ def test_Wavelet3D_PyWt(self):
...
+            x_ad = numpy.vdot(f_p, f)
+            numpy.testing.assert_allclose(x_d, x_ad, rtol=1e-6)
+        print("Undecimated Wavelet 2D adjoint test passes")
+
+    def test_Wavelet_UD_2D_Multichannel(self):
+        """Test the adjoint operation for Undecmated wavelet Transform in
+        multichannel case"""
+        for i in range(self.max_iter):
+            print("Process Wavelet Undecimated test '{0}'...", i)
+            wavelet_op = WaveletUD2(nb_scale=4, multichannel=True, n_cpu=2)
```
zaccharieramzi: Is it executing properly on different cpus in Travis? I mean do we actually have multiple cores?

mri/test/test_wavelet_adjoint.py
```diff
@@ -80,6 +85,40 @@ def test_Wavelet3D_PyWt(self):
             numpy.testing.assert_allclose(x_d, x_ad, rtol=1e-6)
         print(" Wavelet3 adjoint test passes")
 
+    def test_Wavelet_UD_2D(self):
+        """Test the adjoint operation for Undecimated wavelet
+        """
+        for i in range(self.max_iter):
+            print("Process Wavelet Undecimated test '{0}'...", i)
+            wavelet_op = WaveletUD2(nb_scale=4)
+            Img = (numpy.random.randn(self.N, self.N) +
```
zaccharieramzi: nitpicking here, but generally in Python the variables respect snake_case, so this should be `img` (and `i_p`).
More generally, the following variable names are a bit unclear.

mri/reconstruct/linear.py
```diff
@@ -114,13 +116,185 @@ def l2norm(self, shape):
...
+        self.n_cpu = n_cpu
+        self._opt = [
+            '-t{}'.format(self.wavelet_id),
+            '-n{}'.format(self.nb_scale),
+        ]
+        self._has_run = False
+        self.coeffs_shape = None
+
+    def _get_filters(self, shape):
+        """Function to get the Wavelet coefficients of Delta[0][0].
```
zaccharieramzi: Just as a tip for next time, you don't have to document private functions.
Now that it's here let's leave it.

mri/reconstruct/linear.py
chaithyagr: So you would want them as 
```
from joblib import delayed
import numpy as np
from joblib import Parallel
```

Or just that `joblib` import must occur before `numpy`?

mri/reconstruct/linear.py
chaithyagr: _run is an important one that tracks if it has run once, basically ensures we have pulled the transform in with help of ModOpt. It helps to understand code better. 
Although, I agree that from user perspective, it is not needed.

mri/reconstruct/linear.py
chaithyagr: Well its present, without much significance for now, I could as well remove it.

mri/reconstruct/linear.py
chaithyagr: Hmm, interesting, let me generalize this

mri/reconstruct/linear.py
zaccharieramzi: well `joblib` before `numpy` (always the first letter that matters), and `modopt` before `pysap`

mri/reconstruct/linear.py
zaccharieramzi: Yes I guess it's just a comment.

mri/test/test_wavelet_adjoint.py
chaithyagr: Yes, I tried and the test passes! 
I had a submit to check if it works.

I did this experiment after checking here: https://github.com/grosser/parallel_tests/wiki/Distributed-Parallel-Tests-and-Travis-Support

It mentions that : 
> Travis recommends using 2 processes for their 1.5 cores per box.

mri/test/test_wavelet_adjoint.py
chaithyagr: Coding style, while I agree with you, this is how it is currently for all the tests. 
Maybe we fix this in #36 ? 
For now, I feel either we fix all or fix none.

mri/reconstruct/linear.py
chaithyagr: Sure

mri/reconstruct/linear.py
chaithyagr: cool, will change it in next commit

mri/test/test_wavelet_adjoint.py
chaithyagr: Resolving for now.

mri/test/test_wavelet_adjoint.py
chaithyagr: Had an offline discussion, To ensure we have clean check-ins, I am convinced that we make this change now. Will update in next commit

mri/reconstruct/linear.py
chaithyagr: Updated as verbosity level of Parallel. This would be helpful in debug

chaithyagr:
Yes, I'll let @LElgueddari just give a quick look at it. Ill mostly talk off-hand with @AGrigis about this.
More importantly right now to get this in, we would need https://github.com/CEA-COSMIC/ModOpt/pull/64 to be in so that I could point to master branch than my github repository

AGrigis:
I suggest to test against the master banch of pysap and modopt but also from the released version available on pypi.
It can avoid issues with the end user that will probably mix everything.

chaithyagr:
> I suggest to test against the master banch of pysap and modopt but also from the released version available on pypi.
> It can avoid issues with the end user that will probably mix everything.

I agree on this, we could use matrix to test with both cases. However, we would need pysap to export the sparse2D binaries to a location for this, (perhaps in `/.local/pysap/bin` , so that we could do a pip install and point the PATH variable. 
I will open an issue in pysap-mri to update travis, and one in PySAP to ensure we export the binaries.

chaithyagr:
All codes are in. Travis updated to point to CEA-COSMIC master branch of Modopt. 
Last checks by @LElgueddari and we can get this to mainline..

chaithyagr:
Had an offline chat with @LElgueddari , we plan to merge this as there isnt any specific feedback we may require from Loubna for this merge. 

Merging to Master

PR #41, opened 2019-10-03
Add Examples for Cartesian and Non Cartesian 2D MRI reconstructions

chaithyagr:
This addresses:
1) 2 examples in #35  : Cartesian and Non-Cartesian 2D MRI reconstruction
2) The implementation of GenericCost function
3) Uses WaveletUD2 class from #38 for correct adjoint operation


examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
+"""
+Neuroimaging cartesian reconstruction
+=====================================
+
+Author: Chaithya G R
+
+In this tutorial we will reconstruct an MRI image from the sparse kspace
+measurments.
```
zaccharieramzi: typo: measurements

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+
+Import neuroimaging data
+------------------------
+
+We use the toy datasets available in pysap, more specifically a 2D brain slice
+and the acquisition cartesian scheme.
```
zaccharieramzi: cartesian acquisition scheme

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+
+Author: Chaithya G R
+
+In this tutorial we will reconstruct an MRI image from the sparse kspace
+measurments.
+
+Import neuroimaging data
+------------------------
+
+We use the toy datasets available in pysap, more specifically a 2D brain slice
```
zaccharieramzi: we should have somewhere the specification of the brain slice (origin, contrast, FOV, slice thickness, etc...)

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+and the acquisition cartesian scheme.
+"""
+
+# Package import
+from modopt.math.metrics import ssim
+from mri.numerics.reconstruct import sparse_rec_fista
```
zaccharieramzi: try to sort the import alphabetically (`reconstruct` after `fourier`)

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+from mri.numerics.fourier import FFT2
+from mri.numerics.utils import generate_operators
+from mri.numerics.utils import convert_mask_to_locations
+import pysap
+
+# Third party import
+import numpy as np
+
+# Loading input data
+image = pysap.utils.load_image('../../../Data/Pysap_examples/base_image.npy')
```
zaccharieramzi: specify in a TODO that this is just temporary and needs fixing. Also specify it corresponds to the data we will retrospectively undersample

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+
+# View Input
+image.show()
+mask.show()
+
+#############################################################################
+# Generate the kspace
+# -------------------
+#
+# From the 2D brain slice and the acquisition mask, we generate the acquisition
```
zaccharieramzi: a better formulation is that we "retrospectively undersample the k-space using a cartesian acquisition mask"

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+#############################################################################
+# FISTA optimization
+# ------------------
+#
+# We now want to refine the zero order solution using a FISTA optimization.
+# Here no cost function is set, and the optimization will reach the
+# maximum number of iterations.
+
+# Generate operators
+gradient_op, linear_op, prox_op, cost_op = generate_operators(
```
zaccharieramzi: it's a bit misleading to say that no cost function is set, and have `cost_op`, just do 
```python
gradient_op, linear_op, prox_op, _ = generate_operators(
```

and have `None` on line 89.

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+    data=kspace_data,
+    wavelet_name="sym8",
+    samples=kspace_loc,
+    nb_scales=4,
+    mu=2.364e-07,
```
zaccharieramzi: would `2 * 1e-7` work? It's way more readable and concise.
I mean would we still be better than zero-filled recon? If yes then use this and explain how you got to this value in a comment (i.e. grid search).

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+    non_cartesian=False,
+    uniform_data_shape=None,
+    gradient_space="synthesis",
+    padding_mode="periodization")
+
+# Start the FISTA reconstruction
+max_iter = 200
```
zaccharieramzi: have you checked that we are at convergence?

examples/non_cartesian_reconstruction.py
```diff
@@ -0,0 +1,97 @@
...
+
+# Third party import
+from modopt.math.metrics import ssim
+import numpy as np
+
+# Loading input data
+image = pysap.utils.load_image('../../../Data/Pysap_examples/base_image.npy')
+
+# Obtain MRI non-cartesian mask
+radial_mask = get_sample_data("mri-radial-samples")
```
zaccharieramzi: will this currently work?

examples/non_cartesian_reconstruction.py
```diff
@@ -0,0 +1,97 @@
+"""
```
zaccharieramzi: Most of the remarks I made above are to be applied here.

mri/reconstruct/utils.py
```diff
@@ -170,10 +171,13 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
         raise ValueError("At the moment, this functuion only supports 2D "
                          "data.")
     # Define the linear/fourier operators
-    linear_op = WaveletN(
-        nb_scale=nb_scales,
-        wavelet_name=wavelet_name,
-        padding_mode=padding_mode)
+    if wavelet_name != 'UndecimatedBiOrthogonalTransform':
```
zaccharieramzi: what is this?

mri/reconstruct/utils.py
```diff
@@ -207,19 +211,11 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
...
-        cost_op = DualGapCost(
-            linear_op=linear_op,
-            initial_cost=1e6,
-            tolerance=1e-4,
-            cost_interval=1,
-            test_range=4,
-            verbose=0,
-            plot_output=None)
-
+    cost_op = GenericCost(
```
zaccharieramzi: I guess this is the outcome of the whole dual gap conversation right?

mri/reconstruct/utils.py
```diff
@@ -207,19 +211,11 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
             fourier_op=fourier_op)
 
     # Define the proximity dual/primal operator
-    prox_op = SparseThreshold(linear_op, mu, thresh_type="soft")
+    prox_op = SparseThreshold(Identity(), mu, thresh_type="soft")
```
zaccharieramzi: Why has this changed?

examples/cartesian_reconstruction.py
chaithyagr: This does make sense, though I dont know whether we need to do this here, given we will be exporting the data. @philouc could give a better idea on this.

examples/cartesian_reconstruction.py
chaithyagr: Hmm, I dont think I did that detailed sorting. Will get this done

examples/cartesian_reconstruction.py
chaithyagr: Yes, I missed that.

examples/cartesian_reconstruction.py
chaithyagr: Yes. Will update

examples/cartesian_reconstruction.py
chaithyagr: Oops, I will have to update this. I have updated to provide the cost function here. So that is not the right statement

examples/cartesian_reconstruction.py
chaithyagr: Well it surely better, just that this is exactly where I received the maxima in the grid search. 
Yes, I will document on how we got this value. Maybe I will initialize this as a parameter above and add these comments.

However, I think it is better to use the right value of `mu` than a readable value of `mu`.

examples/cartesian_reconstruction.py
chaithyagr: Nope. Will check. I am afraid I need to see how to set up the Modopt for correct convergence checks to prevent manual checks. 
But I do see cost function flattening out at tail in plot

examples/non_cartesian_reconstruction.py
chaithyagr: Yes, resolving

mri/reconstruct/utils.py
chaithyagr: This was the only way I could check for undecimated wavelet and provide a different implementation through Modopt. Any update to make this more scalable would be helpful.

mri/reconstruct/utils.py
chaithyagr: Yes, this does not implement the dual gap cost, it is just the Generic cost which is gradient cost + proximity cost. 
I will add a TODO here as we arent sure if this is what we want always. Maybe, we can add a parameter to choose between DualGap cost and this

examples/cartesian_reconstruction.py
zaccharieramzi: I don't think so. This "right value" of `mu` was grid searched but it doesn't mean that it's the "right one". It wouldn't work in other situations, with a different noise, so it's better to just be readable, explain how and not have people wonder why there is such an exact value here.

examples/cartesian_reconstruction.py
zaccharieramzi: Ok then it's fine, for now manual checking will do

mri/reconstruct/utils.py
zaccharieramzi: you probably have the list of names corresponding to `pywt` or isap somewhere in `pysap`. Use that to determine what constructor to chose.

Found it: `pysap.wavelist()` (I just did `dir(pysap)`)

mri/reconstruct/utils.py
chaithyagr: Long story:
Generic cost has a calculate cost as follows:
 ```
        if self.analysis:
            cost = self.gradient_op.cost(x_new) + self.prox_op.cost(
                self.linear_op.op(x_new))
        else:
            cost = self.gradient_op.cost(x_new) + self.prox_op.cost(
                x_new)
```

Now you would notice that the cost call of proximity for synthesis formulation is with `x_new` which is the wavelet coefficient. If I place Sparse Threshold as my proximity with the same linear operator, we tend to find cost by trying to obtain the wavelet coefficients from `x_new` . This is not right. 
So the proximity is actually Sparsity of wavelet coefficients directly (hence the identity)

File: mri/reconstruct/utils.py (hunk 13 above)
Comment (zaccharieramzi) [reply mri/reconstruct/utils.py]: and definitely add this in the PR description, it's another change compared to just adding examples.

Like the possibility to have sparse2d wavelets with the correct adjoint (even if it's a temporary fix).

File: examples/cartesian_reconstruction.py (hunk 1 above)
Comment (chaithyagr) [reply examples/cartesian_reconstruction.py]: Thanks, on it

File: examples/cartesian_reconstruction.py (hunk 8 above)
Comment (chaithyagr) [reply examples/cartesian_reconstruction.py]: I mean from viewpoint of results, we would want it to be as good as possible. 
With this in mind, I meant the `right_value` as the value that gives the best reconstruction. 
I am fine to make it readable if the reconstruction results and SSIM is not way off.

File: mri/reconstruct/utils.py (hunk 12 above)
Comment (chaithyagr) [reply mri/reconstruct/utils.py]: No, theres a catch. 
Modopt works by sending wavelet ID. 
So in my opinion the best solution under given constraints that we dont modify pySAP or sparse2D or Modopt would be:
* Create a map between the wavelet names from pysap.wavelist and wavelet_id (see help in `mr_transform`)
* Allow user to choose between the Modopt or pysap implementation. As there is not clear reason why the WaveletUD would not support for decimated wavelets also.


For future and best fix:
* Sparse2D needs to provide a way to instance wavelets from wavelet name just like pysap, or it needs to have a way to map it internally.
* Modopt need to call right or map things right.

File: mri/reconstruct/utils.py (hunk 13 above)
Comment (chaithyagr) [reply mri/reconstruct/utils.py]: Yes, Ill update description

File: mri/reconstruct/utils.py (hunk 14 above)
Comment (zaccharieramzi) [reply mri/reconstruct/utils.py]: But why don't we have the `prox_op` in the case of analysis have the `linear_op` ?
Most probably, `GenericCost` should not have to do this distinction.

File: examples/cartesian_reconstruction.py (hunk 8 above)
Comment (zaccharieramzi) [reply examples/cartesian_reconstruction.py]: They shouldn't be with `2 * 1e-7`

File: mri/reconstruct/utils.py (hunk 12 above)
Comment (zaccharieramzi) [reply mri/reconstruct/utils.py]: What you can do is allow the user to specify the name. 
If the name belongs to the list of `pysap` wavelets:
 - if it's `pywt` you implement with `WaveletN`
 - it it's in `isap` you throw a warning and implement with `WaveletN`
If it doesn't it means it belongs to the `mr_transform` ids, so you implement with `WaveletUD2`.
In the example you would just have to specify to which transform you are pointing with an explanation saying it's a temporary fix (and that not all wavelets of `mr_transform` are supported at this point).

File: mri/reconstruct/utils.py (hunk 14 above)
Comment (chaithyagr) [reply mri/reconstruct/utils.py]: I dont think I understood you clearly. 
For the analysis formulation, x_new is actual image domain variable. 

So we could have a generic cost that just implements:
```python
cost = self.gradient_op.cost(x_new) + self.prox_op.cost(x_new)
```

and for `prox_op` have as follows:
```python
if self.analysis:
    prox_op = SparseThreshold(linear_op, mu, thresh_type="soft")
else:
    prox_op = SparseThreshold(Identity(), mu, thresh_type="soft")
```

This would essentially achieve the same thing. But yes, the `generic_cost` as a function would be more "generic"

examples/cartesian_reconstruction.py
chaithyagr: Are you suggesting to have codes with 2 * 1e-7 but the results with 2.364e-07 ? 
I mean I dont know how much would the SSIM change, but would it not kill re-producibility?

mri/reconstruct/utils.py
zaccharieramzi: Yes it's the same thing in the end, but the setting of linear op or not is done at the right time.

chaithyagr: I agree, will resolve this

mri/reconstruct/utils.py
chaithyagr: But then we still need a map to convert names to wavelet IDs , am I right?

examples/cartesian_reconstruction.py
zaccharieramzi: I am suggesting to have codes and results with `2 * 1e-7`

mri/reconstruct/utils.py
zaccharieramzi: no, if you want to use the `ModOpt` you have to provide the wavelet id as the wavelet name.
Since this is a temporary fix it's ok. We can remind somewhere that a call to `mr_transform`'s help will give this map.

chaithyagr: Hmm, dirty fix, but I am ready to go with it for now

chaithyagr: Ahh I think I missed this last time. I dont think I agree with providing `pywt` or `isap` as input parameter.
A simple approach would be:
try:
    WaveletN(wavelet_name=wavelet_name)
except: 
    WaveletUD2(wavelet_id=wavelet_name)
end

I am trying to get rid of this difference between pywt or isap implementation, as the wavelet name clearly distinguishes it. (pywt usually have them as short forms)

For now, Ill fix based on my above proposal, let me know if you have any other points or drawbacks in my approach.

zaccharieramzi: yes I think it's fine for now, since it's just temporary anyway

examples/cartesian_reconstruction.py
chaithyagr: I checked that SSIM is 0.9143 (baseline is 0.82) 
I think we can hold on to 2 * 1e-7 this. Resolving

mri/numerics/reconstruct.py
```diff
@@ -348,19 +348,12 @@ def sparse_rec_condatvu(gradient_op, linear_op, prox_dual_op, cost_op,
...
-        transform_output = linear_op.transform
-    else:
-        linear_op.coeff = opt.y_final
-        transform_output = linear_op.coeff
     if hasattr(cost_op, "cost"):
         costs = cost_op._cost_list
     else:
         costs = None
 
-    return x_final, transform_output, costs, opt.metrics
```
chaithyagr: @LElgueddari , I have removed these as me and @zaccharieramzi feel that there is no use of this. The transform_Output is actually available in linear_op which is an input to this function.
Please correct me if there is any other future purpose of this

LElgueddari: I don't see any objection to this.

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,98 @@
...
+# We now want to refine the zero order solution using a FISTA optimization.
+# The cost function is set to Proximity Cost + Gradient Cost
+
+# Generate operators
+gradient_op, linear_op, prox_op, cost_op = generate_operators(
+    data=kspace_data,
+    wavelet_name="sym8",
+    samples=kspace_loc,
+    nb_scales=4,
+    mu=2 * 1e-7,
```
LElgueddari: How did you set the value of mu?

chaithyagr:
Updated the codes with modification for the test to call generate_operators. There is still some issue with Undecimated Wavelet for optimization tests as they seem o converge, but not in a way that the reconstruction matches the zero-filled Inverse FT.

Also, I noticed that Mallat 79 filters returns NaN at a point. Earlier, with self.AssertTrue, in test, we had a false pass. Now I moved the tests to check with sym8.
There seem to be 2 issues that needs to be opened at this point:
1) Mallat Filters not Working right / Causing NaN
2) Undecimated Wavelet causes convergence which is not very close to 0 filled Inverse FT.

For now, I hope these tests will pass. Adding Zac for a re-review

mri/numerics/reconstruct.py
LElgueddari: I feel like you should return also `y_final` since it is the dual solution and that primal-dual method returns both the primal and the dual. Specifically for Condat Vu method.

examples/cartesian_reconstruction.py
chaithyagr: After a grid search for it. Me and @zaccharieramzi had a discussion on this and how it is not optimal. You can refer to that here: https://github.com/CEA-COSMIC/pysap-mri/pull/41#discussion_r330936408

examples/non_cartesian_reconstruction.py
```diff
@@ -0,0 +1,96 @@
...
+#
+# From the 2D brain slice and the acquisition mask, we retrospectively
+# undersample the k-space using a radial acquisition mask
+# We then reconstruct the zero order solution as a baseline
+
+# Get the locations of the kspace samples and the associated observations
+fourier_op = NFFT(samples=kspace_loc, shape=image.shape)
+kspace_obs = fourier_op.op(image.data)
+
+# Zero order solution
```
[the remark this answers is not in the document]
chaithyagr: Well this is the easiest from coding perspective. Yes, maybe zero filled is a misnomer. Do you have any other suggestions for re-gridding? I would want this approach to be as a function in utils. (even if we use griddata)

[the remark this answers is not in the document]
LElgueddari: I probably suggest having a real gridding step using griddata than having this "trick" which is fairer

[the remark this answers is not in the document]
chaithyagr: Makes sense. Let me give that a try and make it a function in utils.py

chaithyagr:
With this, I would consider this PR to be complete. @zaccharieramzi can you do a final check and once all tests pass merge it to mainline?

examples/non_cartesian_reconstruction.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# We now want to refine the zero order solution using a FISTA optimization.
+# The cost function is set to Proximity Cost + Gradient Cost
+
+# Generate operators
+gradient_op, linear_op, prox_op, cost_op = generate_operators(
+    data=kspace_obs,
+    wavelet_name=24,
+    samples=kspace_loc,
+    mu=6.45e-07,
```
zaccharieramzi: would `6 * 1e-7` work ?

mri/numerics/cost.py
```diff
@@ -13,3 +13,4 @@
 
 # Package import
 from mri.reconstruct.cost import DualGapCost
+from mri.reconstruct.cost import GenericCost
```
zaccharieramzi: I think importing the 2 on the same line is ok

mri/test/test_optimizer.py
```diff
@@ -37,8 +32,14 @@ def setUp(self):
...
             [im.data.shape for im in self.images]))
         self.mask = get_sample_data("mri-mask").data
-        self.names = ["MallatWaveletTransform79Filters"]
-        print("[info] Found {0} transformations.".format(len(self.names)))
+        # Test a wide variety of linear operators :
+        # From WaveletN
+        self.decimated_wavelets = ['sym8']
+        # From WaveletUD2, tested only for analysis formulation
+        self.undecimated_wavelets = [24]
+        # , 24] TODO debug and add this for testing
```
zaccharieramzi: I don't understand this comment

chaithyagr: Ahh, this was for fixing above described assumed bug for waveletUD. Now it doesn't make sense. Will remove it.

mri/numerics/cost.py
chaithyagr: Cool, will update

examples/non_cartesian_reconstruction.py
chaithyagr: Will check and report change in SSIM here. If it is not significant, Ill make the change

chaithyagr: SSIM is now, 0.9210 it was approx same, changed.

examples/cartesian_reconstruction.py
zaccharieramzi: Let's have that as an issue and resolve this.

mri/numerics/reconstruct.py
```diff
@@ -348,19 +344,13 @@ def sparse_rec_condatvu(gradient_op, linear_op, prox_dual_op, cost_op,
...
-        linear_op.coeff = opt.y_final
-        transform_output = linear_op.coeff
+    y_final = opt.y_final
     if hasattr(cost_op, "cost"):
         costs = cost_op._cost_list
     else:
         costs = None
 
-    return x_final, transform_output, costs, opt.metrics
+    return x_final, costs, opt.metrics, y_final
```
zaccharieramzi: This output should be in the doc

examples/cartesian_reconstruction.py
chaithyagr: Yes. Resolving it here

mri/numerics/reconstruct.py
chaithyagr: On it.

chaithyagr:
All checks are done. @zaccharieramzi my last commit just changed doc, so we don't have to wait for Travis, can you please merge this to master?

PR #43, opened 2019-10-10
[VERY SMALL] Use get_sample_data to load example data

chaithyagr:
In this example, we use get_sample_data top load pysap-mri example data


philouc:
This API is perfect guys! Thanks for figuring out this issue.

Issue #42, opened 2019-10-04
Add tests for POGM optimizer


chaithyagr:
This issue tracks writing 2 tests for POGM: Cartesian and Non-Cartesian cases. 

We will have linear operators with mu=0, and ensure that we have converged to adjoint of fourier operator

chaithyagr:
I have a PR ready for this in https://github.com/chaithyagr/pysap-mri/tree/pogm_test
However, we will have to wait for #41 to hit to mainline to get this in.

LElgueddari:
Any updates on this PR?

PR #44, opened 2019-10-16
Add tests for POGM

chaithyagr:
This PR adds tests for POGM. Tracked in #42

### DIFF
--- a/mri/numerics/reconstruct.py
+++ b/mri/numerics/reconstruct.py
@@ -443,7 +443,7 @@ def sparse_rec_pogm(gradient_op, linear_op, prox_op, cost_op=None,
         print("Done.")
         print("Execution time: ", end - start, " seconds")
         print("-" * 40)
-    x_final = opt.x_final
+    x_final = linear_op.adj_op(opt.x_final)
     metrics = opt.metrics
 
     if hasattr(cost_op, "cost"):
--- a/mri/test/test_optimizer.py
+++ b/mri/test/test_optimizer.py
@@ -175,6 +175,46 @@ class TestOptimizer(unittest.TestCase):
                     np.testing.assert_allclose(x_final, np.fft.ifftshift(
                         fourier_0.adj_op(data_0)))
 
+    def test_reconstruction_pogm_nfft2(self):
+        """ Test reconstruction with POGM.
+        """
+        print("Process test NFFT2 FISTA::")
+        for image in self.images:
+            fourier = NFFT(samples=convert_mask_to_locations(
+                                            self.mask),
+                           shape=image.shape)
+            data = fourier.op(image.data)
+            print("Process test with image '{0}'...".format(
+                image.metadata["path"]))
+            for nb_scale in self.nb_scales:
+                print("- Number of scales: {0}".format(nb_scale))
+                for name in self.decimated_wavelets:
+                    print("    Transform: {0}".format(name))
+                    gradient_op, linear_op, prox_op, cost_op = \
+                        generate_operators(
+                            data=data,
+                            wavelet_name=name,
+                            samples=convert_mask_to_locations(self.mask),
+                            mu=0,
+                            nb_scales=4,
+                            non_cartesian=True,
+                            uniform_data_shape=image.shape,
+                            gradient_space="synthesis")
+                    x_final, costs, _ = sparse_rec_fista(
+                        gradient_op=gradient_op,
+                        linear_op=linear_op,
+                        prox_op=prox_op,
+                        cost_op=cost_op,
+                        lambda_init=1.0,
+                        max_nb_of_iter=self.nb_iter,
+                        verbose=0)
+                    fourier_0 = FFT2(samples=convert_mask_to_locations(
+                                            fftshift(self.mask)),
+                                     shape=image.shape)
+                    data_0 = fourier_0.op(np.fft.fftshift(image.data))
+                    np.testing.assert_allclose(x_final, np.fft.ifftshift(
+                        fourier_0.adj_op(data_0)))
+
     def test_reconstruction_condat_vu_nfft2(self):
         """ Test all the registered transformations.
         """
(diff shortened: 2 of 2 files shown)


mri/numerics/reconstruct.py
```diff
@@ -443,7 +443,7 @@ def sparse_rec_pogm(gradient_op, linear_op, prox_op, cost_op=None,
         print("Done.")
         print("Execution time: ", end - start, " seconds")
         print("-" * 40)
-    x_final = opt.x_final
+    x_final = linear_op.adj_op(opt.x_final)
```
chaithyagr: This is done to follow the same norm as other optimizers. We want to ensure that we return Images in the end.

zaccharieramzi: I really dislike this as the `adj_op` is not necessarily an inverse (or reconstruction) operation. I don't know how it's done elsewhere, but we should let the user change the outcome the way they want.

mri/test/test_optimizer.py
```diff
@@ -17,6 +17,7 @@
 from mri.reconstruct.fourier import FFT2, NFFT
 from mri.numerics.reconstruct import sparse_rec_fista
 from mri.numerics.reconstruct import sparse_rec_condatvu
+from mri.numerics.reconstruct import sparse_rec_pogm
```
zaccharieramzi: Maybe we can have all these imports on the same line.

mri/test/test_optimizer.py
```diff
@@ -84,6 +85,47 @@ def test_reconstruction_fista_fft2(self):
                     np.testing.assert_allclose(x_final, np.fft.ifftshift(
                         fourier_0.adj_op(data_0)))
 
+    def test_reconstruction_pogm_fft2(self):
+        """ Test all the registered transformations.
+        """
+        print("Process test FFT2 POGM::")
+        for image in self.images:
+            fourier = FFT2(samples=convert_mask_to_locations(
+                                            fftshift(self.mask)),
```
zaccharieramzi: Can you explain why we have to do this?

mri/test/test_optimizer.py
```diff
@@ -84,6 +85,47 @@ def test_reconstruction_fista_fft2(self):
...
+                        prox_op=prox_op,
+                        cost_op=cost_op,
+                        max_nb_of_iter=self.nb_iter,
+                        verbose=0)
+                    fourier_0 = FFT2(samples=convert_mask_to_locations(
+                                            fftshift(self.mask)),
+                                     shape=image.shape)
+                    data_0 = fourier_0.op(np.fft.fftshift(image.data))
+                    np.testing.assert_allclose(x_final, np.fft.ifftshift(
+                        fourier_0.adj_op(data_0)))
```
zaccharieramzi: All these ffshifts are making me nervous. I don't understand why they are not handled directly in the FFT objects we have.

mri/test/test_optimizer.py
chaithyagr: Well, PEP8 would still need to move to a different line, this just looks cleaner to me. Let me know if this needs to be changed still, I am fine with it also

mri/test/test_optimizer.py
chaithyagr: Well this is something to be fixed, in another PR. FFT class right now works on shifted data, while all other Fourier classes (NFFT, NUFFT etc) works directly on mask. 
I would really like FFT to change its behaviour, but I would not do this in this PR. Maybe I will get an issue up for it. Will resolve this after referencing the issue here

mri/test/test_optimizer.py
chaithyagr: Resolving as a duplicate of above

mri/test/test_optimizer.py
zaccharieramzi: Yes definitely don't do it here, but it seems weird to me. I had never noticed but fair enough.

mri/numerics/reconstruct.py
LElgueddari: For a more general perspective, the sparse_rec_* is supposed to give the image solution to the reconstruction problem, if we assume that this is the goal of the function than we should output the image. This is for a general perspective, I think that to handle this in a correct manner, `linar_op` should have both an adjoint operator and an inverse operator, i.e. for the simplest case -- i.e. orthogonal wavelet transform, handling correctly the boundaries --  the adjoint and the inverse should be the same whereas for more complicated  transforms maybe we should implement a dedicated inverse operation.

mri/test/test_optimizer.py
chaithyagr: Well I planned to move to that itself, just didnt know how clean it was. Will get an update

mri/test/test_optimizer.py
LElgueddari: Most of the time when you compute your mask you are putting the DC frequency at the middle of the mask, however, the FFT operator does not compute it in the middle but rather in the boundaries, therefore you either need to add an `fftshift` operation directly on the mask or for each `FFT` operations  this was the easiest solution.

chaithyagr: Nah, you might have come across this even in cartesian example:

https://github.com/CEA-COSMIC/pysap-mri/blob/5bb4d2ffc95997fa180d6057ac7e3618c16e99e0/examples/cartesian_reconstruction.py#L48-L49


So there will surely be a bigger impact of this change, although it is a right change

chaithyagr: #45 created. Resolving here

mri/numerics/reconstruct.py
chaithyagr: I agree with Loubna, resolving here. We could instance an issue to obtain the reconstruction operator in both WaveletN and WaveletUD classes.

zaccharieramzi: Yes sorry I was misreading, I thought this was the algorithm. This is def fine.

mri/test/test_optimizer.py
zaccharieramzi: You didn't look at the SO answer though haha, nv mind it's ok like this

PR #47, opened 2019-10-18
[URGENT] Add verbosity for costOp, make default False

chaithyagr:
This is to ensure we dont calculate and print out cost at every iteration step.
Refer to https://github.com/CEA-COSMIC/pysap/pull/111#discussion_r336394197


chaithyagr:
@zaccharieramzi can you please merge this as it has passed?

PR #50, opened 2019-10-23
Fixing installation issues

LElgueddari:
This fixes the installation issues of scikit-image for the version of python = 3.5


setup.py
```diff
@@ -48,8 +48,7 @@
     setup_requires=["numpy",
                     "scipy",
                     "cython",
-                    "pytest-runner",
-                    "scikit-image"],
+                    "pytest-runner"],
```
chaithyagr: Why did you remove scikit-image from here? 
If I am right, the `pip install` would now not install scikit-image. 
Did you verify that? 
Going forward, we would want 2 versions of tests, one working with python setup.py install and other with pip install. Maybe we can cover this in a separate issue where we are adding examples to travis / circleCI

LElgueddari: Ther is a piece of redundant information basically we defined the setup_requires line 33 and could just use it again inside the setup
```suggestion
setup_requires = ["numpy", "scipy", "cython", "pytest-runner", "scikit-image"]

pip_main(['install'] + setup_requires)

setup(
    name="pysap-mri",
    description="Python Sparse data Analysis Package external MRI plugin.",
    long_description="Python Sparse data Analysis Package external MRI plugin.",
    license="CeCILL-B",
    classifiers="CLASSIFIERS",
    author=AUTHOR,
    author_email="XXX",
    version="0.1.1",
    url="https://github.com/CEA-COSMIC/pysap-mri",
    packages=find_packages(),
    setup_requires=setup_requires,
    install_requires=["scipy",
                      "numpy",
                      "scikit-learn",
                      "progressbar2",
                      "joblib",
                      "psutil",
                      "pynfft@git+https://github.com/ghisvail/pyNFFT@master"],
    dependency_links=['https://github.com/ghisvail/pyNFFT/tarball/master#egg=pynfft-1.3.0'],
    tests_require=['pytest>=5.0.1', 'pytest-cov>=2.7.1', 'pytest-pep8'],
    platforms="OS Independent"
)
```
If we do this why would we have numpy installd both in the install_requires, I think it needs a bit of cleaning here

chaithyagr: Yes, that can be removed. Maybe this PR is perfect for it. This cleanup, would also ensure that you will not miss out on scikit image. I feel its best solution..

chaithyagr: Latest commit resolves this..

PR #51, opened 2019-10-23
FFT shift and support for multi-coil 3D FFT

LElgueddari:
This PR solves :
- shift issue when using cartesian sampling. Hence Cartesian and non-Cartesian sampling have the same convention (k_space center == DC)
- Multiple coils can be used as it is the case for nonCartesianFT
- Multidimensional FFT operation can be made

This solves #8


zaccharieramzi:
I think this PR does more than what you said right? I see also a name change related to the fact that we accept now multiple coils in the FFT op.

mri/reconstruct/utils.py
```diff
@@ -162,7 +162,7 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
     # Local imports
     from mri.numerics.cost import GenericCost
     from mri.numerics.linear import WaveletN, WaveletUD2
-    from mri.numerics.fourier import FFT2
+    from mri.numerics.fourier import FFT
```
chaithyagr: https://github.com/CEA-COSMIC/pysap-mri/blob/d1ac8ca10852d922f37bab51d56d6801bd6fde06/mri/reconstruct/utils.py#L181-L183

You may have to change this also. We will now start supporting the function

LElgueddari: Ok I'll do it,
We should also probably add an example with the cartesian 3D reconstruction with or without sensitivity maps estimation

chaithyagr: Maybe add it as a part of this PR itself?

LElgueddari: Working on it

examples/non_cartesian_reconstruction_stack_3d.py
```diff
@@ -30,7 +30,7 @@
 
 # Loading input data
 image = get_sample_data('3d-pmri')
-image = pysap.Image(data=np.sum(image.data, axis=0))
+image = pysap.Image(data=np.sqrt(np.sum(np.abs(image.data), axis=0)))
```
chaithyagr: I am a bit confused here. Is it square root of sum of squares? Or just square root of absolute sum? Somehow latter does not sound right to me (the way it is coded)

examples/cartesian_reconstruction_3d.py
```diff
@@ -0,0 +1,99 @@
...
+# Third party import
+import numpy as np
+
+# Loading input data and convert it into a single channel using Sum-Of-Squares
+image = get_sample_data('3d-pmri')
+image.data = np.sqrt(np.sum(np.abs(image.data)**2, axis=0))
+
+# Obtain K-Space Cartesian Mask
+mask = get_sample_data("2d-poisson-disk-mask")
+mask.data = np.repeat(np.expand_dims(mask.data, axis=-1), 160, axis=-1)
```
chaithyagr: Can we try and save the complete mask directly? I feel it would be cleaner code than having even a part of mask preparation in the example. I feel its good to follow the same flow as is done in other examples. Also, the example must be, in my opinion all about how to reconstruct.

mri/reconstruct/fourier.py
```diff
@@ -70,66 +70,105 @@ def adj_op(self, x):
...
+        if self.n_coils == 1:
+            return self._mask * np.fft.ifftshift(np.fft.fftn(
+                                    np.fft.fftshift(img), norm="ortho"))
+        else:
+            if self.n_coils > 1 and self.n_coils != img.shape[0]:
+                raise ValueError("The number of coils parameter is not equal"
+                                 "to the actual number of coils, the data must"
+                                 "be reshaped as [n_coils, Nx, Ny, Nz]")
+            else:
+                return np.asarray([self._mask * np.fft.ifftshift(np.fft.fftn(
```
chaithyagr: Do we have speedup if we use joblib and run on multiple CPUs? Just checking. Unlike NFFT, if this gives significant speedup, it would be good to expose it to user.

LElgueddari: Make it as an enhancement

examples/non_cartesian_reconstruction_stack_3d.py
```diff
@@ -30,7 +30,7 @@
 
 # Loading input data
 image = get_sample_data('3d-pmri')
-image = pysap.Image(data=np.sum(image.data, axis=0))
+image = pysap.Image(data=np.sqrt(np.sum(np.abs(image.data)**2, axis=0)))
```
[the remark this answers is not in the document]
LElgueddari: Yes we could although I am not sure that for tutorial/example purposes it is  great since we compute the SOS as it has been suggested by [Roemer](https://onlinelibrary.wiley.com/doi/epdf/10.1002/mrm.1910160203)

examples/cartesian_reconstruction_3d.py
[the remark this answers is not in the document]
LElgueddari: Its the Z-dimension changing it right now

examples/non_cartesian_reconstruction_stack_3d.py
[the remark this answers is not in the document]
zaccharieramzi: what I am saying is that this formula is I think equivalent to :
`np.linalg.norm(image.data, axis=0)` no ?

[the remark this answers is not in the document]
LElgueddari: Yes it is, but having it explicitly make it more explicitly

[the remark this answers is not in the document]
zaccharieramzi: ok sure

PR #52, opened 2019-11-10
Code Refactoring

chaithyagr:
This PR does code refactoring based on : https://github.com/CEA-COSMIC/pysap-mri/issues/36#issuecomment-550997255

This will be a big PR and is currently under progress. The work will be done by me and @LElgueddari .

Review will be done by @zaccharieramzi . Adding @zaccharieramzi right now to start off review..

PR #53, opened 2019-11-12
[SHORT][Code Refactoring] Move linear and fourier Operators

chaithyagr:
This is first PR in a series of PR for carrying out #36 . It is first part of #52


mri/operators/fourier/utils.py
```diff
@@ -0,0 +1,291 @@
...
+# the CEA-CNRS-INRIA. Refer to the LICENSE file or to
+# http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.html
+# for details.
+##########################################################################
+
+"""
+Common tools for MRI image reconstruction.
+"""
+
+
```
chaithyagr: This file just holds the generate operators from previous file. I moved the previous file to mri/operators/fourier/utils as they are all utils for fourier.
@zaccharieramzi  You ideally dont have to review this file.

chaithyagr:
@zaccharieramzi you can review this now.

mri/operators/fourier/non_cartesian.py
```diff
@@ -0,0 +1,706 @@
+# -*- coding: utf-8 -*-
```
zaccharieramzi: I think this file is the exact copy of the one above no?

mri/operators/utils/__init__.py
```diff
@@ -7,10 +7,10 @@
...
+""" This module defines the common operators.
 """
 
-# Package import
-from mri.reconstruct.linear import WaveletN
-from mri.reconstruct.linear import WaveletUD2
+# Fourier utils
+from ..fourier.utils import convert_mask_to_locations, \
+    convert_locations_to_mask, normalize_frequency_locations, \
+    get_stacks_fourier
```
zaccharieramzi: Why not also link to linear utils?

mri/operators/fourier/utils.py
zaccharieramzi: Yes, next time don't forget to use `git mv` if possible.

mri/operators/_base.py
```diff
@@ -0,0 +1,40 @@
+# #############################################################################
```
zaccharieramzi: As discussed, we don't want to have `_base` but `base`.

mri/operators/linear/wavelet.py
```diff
@@ -13,6 +13,7 @@
 
 
 # Package import
+from .._base import OperatorBase
```
zaccharieramzi: Watch out for the name change in `base`.

mri/test/test_optimizer.py
```diff
@@ -48,8 +48,8 @@ def test_reconstruction_fista_fft2(self):
         print("Process test FFT FISTA::")
         for image in self.images:
             fourier = FFT(samples=convert_mask_to_locations(
-                                            fftshift(self.mask)),
```
zaccharieramzi: I don't know what these whitelines changes are.
Ideally we should avoid those (even if can be hidden in github review) because they clutter unnecessarily the git history.
When having to write on multiple lines, the best way to do it to avoid having to make whiteline changes is:
```python
var = long_function_call(
	var1=var1,
	var2=var2,
	var3=var3,
)
```
This will limit at best the number of unnecessary changes (whitelines and commas).

mri/operators/_base.py
```diff
@@ -0,0 +1,40 @@
...
+#  pySAP - Copyright (C) CEA, 2017 - 2018                                     #
+#  Distributed under the terms of the CeCILL-B license,                       #
+#  as published by the CEA-CNRS-INRIA. Refer to the LICENSE file or to        #
+#  http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.html for details.   #
+# #############################################################################
+
+
+class OperatorBase(object):
+    """ Base Operator class. Every operator inherits from this class,
```
zaccharieramzi: every linear operator right?

mri/operators/_base.py
```diff
@@ -0,0 +1,40 @@
...
+    to ensure that we have all the functions rightly implemented
+    as required by Modopt
+    """
+
+    def op(self, img):
```
zaccharieramzi: not necessarily on an image.

mri/operators/_base.py
```diff
@@ -0,0 +1,40 @@
...
+        Returns
+        -------
+        result: np.ndarray
+            operator transform of the input.
+        """
+        raise NotImplementedError("'op' is an abstract method.")
+
+    def adj_op(self, x):
+        """ This method calculates adjoint operator transform of real or
+        complex sequence.
```
zaccharieramzi: not necessarily a sequence. I don't think you need to specify that it can be applied to complex-valued or real-valued data.

mri/operators/utils/__init__.py
chaithyagr: Missed it, cause as of now we dont use linear utils. But will add them.

mri/operators/_base.py
chaithyagr: yes on it

mri/operators/linear/wavelet.py
chaithyagr: yup

mri/test/test_optimizer.py
chaithyagr: Yes, but I will not change them here as this file will change fully in future. (We will use reconstructors)

mri/operators/_base.py
chaithyagr: Every operator, including fourier. Missed on removing fourier base

mri/operators/_base.py
chaithyagr: Will make it general. It is not necessarrily anything here. Its a base template

mri/operators/_base.py
chaithyagr: Again it is a template.

mri/operators/fourier/utils.py
chaithyagr: I did and thats what caused this. What I needed was git copy . Which I dont think exists

mri/operators/utils/__init__.py
zaccharieramzi: I don't think you need to comment these imports, the names are pretty self-explanatory.

mri/operators/base.py
```diff
@@ -0,0 +1,40 @@
...
+# #############################################################################
+
+
+class OperatorBase(object):
+    """ Base Operator class. Every operator inherits from this class,
+    to ensure that we have all the functions rightly implemented
+    as required by Modopt
+    """
+
+    def op(self, input):
```
zaccharieramzi: `input` is protected in Python. I think `x` or `data` are fine here.

mri/operators/base.py
```diff
@@ -0,0 +1,40 @@
+# #############################################################################
+#  pySAP - Copyright (C) CEA, 2017 - 2018                                     #
+#  Distributed under the terms of the CeCILL-B license,                       #
+#  as published by the CEA-CNRS-INRIA. Refer to the LICENSE file or to        #
+#  http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.html for details.   #
+# #############################################################################
+
+
+class OperatorBase(object):
+    """ Base Operator class. Every operator inherits from this class,
```
zaccharieramzi: I don't know why you resolved (let me resolve the conversations I start for the next ones), but here I think we should specify that we are talking about linear operators (even if the notion of adjoint is larger than that).

mri/operators/base.py
```diff
@@ -0,0 +1,40 @@
...
+        Returns
+        -------
+        result: np.ndarray
+            operator transform of the input.
+        """
+        raise NotImplementedError("'op' is an abstract method.")
+
+    def adj_op(self, x):
+        """ This method calculates adjoint operator transform of real or
+        complex data.
```
zaccharieramzi: Again here I think this specification is unnecessary.

mri/operators/utils/__init__.py
chaithyagr: Well now that they are in, I think I will keep them there. Anyway it creates distinction in code lines for future.

mri/operators/base.py
chaithyagr: Oh damn I forgot, yes on it.

mri/operators/base.py
chaithyagr: Ahh, Im sorry, I missed that this had you to answer.
Hmm makes sense..

mri/operators/base.py
chaithyagr: Well, its just an abstract class, so it doesnt really matter, but again I think ill keep it there now that it is there.

zaccharieramzi: Even if it's an abstract class we still want to provide accurate and efficient doc. Here this makes it seem like there is an additional catch.

mri/operators/fourier/utils.py
```diff
@@ -272,8 +272,8 @@ def get_stacks_fourier(kspace_loc):
     sort_pos = np.lexsort(tuple(kspace_loc[:, i]
                                 for i in np.arange(3)))
     kspace_loc = kspace_loc[sort_pos]
-    first_stack_len = np.size(np.where(kspace_loc[:, 2]
-                                       == np.min(kspace_loc[:, 2])))
+    first_stack_len = np.size(np.where(
+        kspace_loc[:, 2] == np.min(kspace_loc[:, 2])))
```
zaccharieramzi: Just as a nitpicking (don't need to correct here), the best way to avoid future unnecessary changes is :
```python
first_stack_len = np.size(np.where(
        kspace_loc[:, 2] == np.min(kspace_loc[:, 2]),
))
```

mri/operators/fourier/utils.py
zaccharieramzi: Caused what?

mri/operators/fourier/utils.py
chaithyagr: Yes. Duly noted, this will change in future anyway.

chaithyagr:
@zaccharieramzi , You can merge this to mainline once everything pass. I will rebase my next set of codes and start a PR immediately/

PR #54, opened 2019-11-13
[SHORT][Code Refactoring] Move Gradient operators and n_coils compat in linear operators

chaithyagr:
This is part 2 of #36


mri/operators/__init__.py
```diff
@@ -17,3 +17,7 @@
 # Linear operators
 from .linear.wavelet import WaveletN, WaveletUD2
 from .linear.dictionary import DictionaryLearning
+
+# Gradient operators
```
zaccharieramzi: Again, I think the import names are self explanatory, I am sorry I let that slip the first time, but I don't think we need those extra comments (see this [SO answer](https://stackoverflow.com/a/7533090/4332585) for example).

mri/operators/linear/wavelet.py
```diff
@@ -28,21 +28,31 @@ class WaveletN(OperatorBase):
...
         """ Initialize the 'WaveletN' class.
-
         Parameters
         ----------
         wavelet_name: str
             the wavelet name to be used during the decomposition.
         nb_scales: int, default 4
             the number of scales in the decomposition.
+        n_coils: int, default 1
+            the number of coils for calibrationless reconstruction
```
zaccharieramzi: typo calibrationless -> multichannel

mri/operators/linear/test_wavelet_adjoint.py
```diff
@@ -107,7 +107,8 @@ def test_Wavelet_UD_2D_Multichannel(self):
         multichannel case"""
         for i in range(self.max_iter):
             print("Process Wavelet Undecimated test '{0}'...", i)
-            wavelet_op = WaveletUD2(nb_scale=4, multichannel=True, n_cpu=2)
+            wavelet_op = WaveletUD2(nb_scale=4, n_coils=self.num_channels,
```
zaccharieramzi: Try to use my technique to avoid future unnecessary changes

mri/operators/utils/__init__.py
```diff
@@ -18,3 +18,6 @@
 # Linear utils
 from ..linear.utils import extract_patches_from_2d_images, min_max_normalize, \
     learn_dictionary
+
+# Gradient utils
```
zaccharieramzi: Same remark as above regarding import comments

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,61 @@
...
+from .utils import check_lipschitz_cst
+
+# Third party import
+import numpy as np
+from modopt.math.matrix import PowerMethod
+from modopt.opt.gradient import GradBasic
+
+
+class GradBaseMRI(GradBasic):
+    """ Gradient Analysis class for single channel
```
zaccharieramzi: I don't think this is the correct doc

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,61 @@
...
+            a function handle that implements Mt X
+        shape : tuple
+            shape of transformed data
+        lipschitz_cst : int default None
+            The lipschitz constant for for given operator.
+            If not specified this is calculated using PowerMethod
+        max_iter_spec_rad : int default 10
+            Number of iterations to calculate the lipschitz constant
+        num_check_lips : int default 10
+            Number of iterations to check if lipschitz constant is correct
```
zaccharieramzi: Try to doc `verbose` as well

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,165 @@
...
+##########################################################################
+# pySAP - Copyright (C) CEA, 2017 - 2018
+# Distributed under the terms of the CeCILL-B license, as published by
+# the CEA-CNRS-INRIA. Refer to the LICENSE file or to
+# http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.html
+# for details.
+##########################################################################
+
+"""
+This module contains classes for defining algorithm operators and gradients.
```
zaccharieramzi: Isn't it just gradients?

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,165 @@
...
+                                               (fourier_op.n_coils,
+                                                *fourier_op.shape),
+                                               verbose=verbose,
+                                               **kwargs)
+        else:
+            super(GradAnalysis, self).__init__(data, fourier_op.op,
+                                               fourier_op.adj_op,
+                                               fourier_op.shape,
+                                               verbose=verbose,
+                                               **kwargs)
```
zaccharieramzi: Why not:
```python
if fourier_op.n_coils != 1:
	data_shape = (fourier_op.n_coils, *fourier_op.shape)
else:
	data_shape = fourier_op.shape
super(GradAnalysis, self).__init__(
	data, 
	fourier_op.op,
	fourier_op.adj_op,
    fourier_op.shape,
    verbose=verbose,
    **kwargs,
)
```

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class GradBaseMRI(GradBasic):
+    """ Gradient Analysis class for single channel
+        Parameters
+        ----------
+        data: np.ndarray
+            input data array.
+        operator :
+            a function handle that implements M X
```
zaccharieramzi: what is `M X`? (same question as for under)

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,61 @@
...
+        trans_operator :
+            a function handle that implements Mt X
+        shape : tuple
+            shape of transformed data
```
zaccharieramzi: What does "transformed data" mean?

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,165 @@
...
+    def _trans_op_method(self, coeff):
+        data_per_ch = self.fourier_op.adj_op(coeff)
+        return np.sum(data_per_ch * np.conjugate(self.Smaps), axis=0)
+
+
+class GradSelfCalibrationSynthesis(GradBaseMRI):
+    """ Gradient Synthesis class for parallel MR reconstruction based on the
+    coil sensitivity profile.
+    This class defines the grad operators for:
+            (1/2) * sum(||Ft Psi_t Sl Alpha - yl||^2_2,l)
```
zaccharieramzi: Isn't it `(1/2) * sum(||Ft Sl Psi_t Alpha - yl||^2_2,l)`?

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,165 @@
...
+    This class defines the grad operators for:
+            (1/2) * sum(||Ft Sl x - yl||^2_2,l)
+    Attributes
+    ----------
+    data: np.ndarray
+        input observed data array.
+    fourier_op: instance
+        a Fourier operator instance.
+    Smaps: np.ndarray
+        Coil sensitivity profile [L, Nx, Ny, Nz]
```
zaccharieramzi: Does it have to be 3D?

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,165 @@
...
+    Attributes
+    ----------
+    data: np.ndarray
+        input observed data array.
+    fourier_op: instance
+        a Fourier operator instance.
+    linear_op: instance
+        a linear operator instance.
+    Smaps: np.ndarray
+        Coil sensitivity profile [L, Nx, Ny, Nz]
```
zaccharieramzi: same question as above

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,165 @@
...
+class GradSelfCalibrationSynthesis(GradBaseMRI):
+    """ Gradient Synthesis class for parallel MR reconstruction based on the
+    coil sensitivity profile.
+    This class defines the grad operators for:
+            (1/2) * sum(||Ft Psi_t Sl Alpha - yl||^2_2,l)
+    Attributes
+    ----------
+    data: np.ndarray
+        input observed data array.
+    fourier_op: instance
```
zaccharieramzi: `instance` is not really a type. Specify the class it implements (same for all the others)

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,165 @@
...
+    def _op_method(self, coeff):
+        image = self.linear_op.adj_op(coeff)
+        image_per_ch = np.asarray([image * self.Smaps[ch]
+                                   for ch in range(self.Smaps.shape[0])])
+        return self.fourier_op.op(image_per_ch)
+
+    def _trans_op_method(self, data):
+        data_per_ch = self.fourier_op.adj_op(data)
+        image_recon = np.sum(data_per_ch * np.conjugate(self.Smaps), axis=0)
+        return self.linear_op.op(image_recon)
```
zaccharieramzi: How did you check that the rewriting of the methods is correct (for the pMRI gradients)?

Ideally we would do the rewriting in a different PR.

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,61 @@
...
+            self.spec_rad = lipschitz_cst
+            self.inv_spec_rad = 1.0 / self.spec_rad
+        else:
+            calc_lips = PowerMethod(self.trans_op_op, shape,
+                                    data_type=np.complex, auto_run=False)
+            calc_lips.get_spec_rad(extra_factor=1.1,
+                                   max_iter=max_iter_spec_rad)
+            self.spec_rad = calc_lips.spec_rad
+            self.inv_spec_rad = calc_lips.inv_spec_rad
+        if num_check_lips > 0:
```
zaccharieramzi: Why do we need this? This might be more a question for @LElgueddari

mri/operators/linear/wavelet.py
```diff
@@ -57,60 +67,89 @@ def get_coeff(self):
     def set_coeff(self, coeffs):
         self.transform.analysis_data = coeffs
 
+    def _op(self, data):
+        if isinstance(data, np.ndarray):
+            data = pysap.Image(data=data)
+        self.transform.data = data
+        self.transform.analysis()
+        coeffs, coeffs_shape = flatten(self.transform.analysis_data)
```
zaccharieramzi: Maybe this question is not for now but for the refactoring in general: do we really need to flatten the coefficients?

mri/operators/linear/wavelet.py
```diff
@@ -57,60 +67,89 @@ def get_coeff(self):
...
-        self.transform.data = data
-        self.transform.analysis()
-        coeffs, self.coeffs_shape = flatten(self.transform.analysis_data)
+        if self.n_coils > 1:
+            coeffs, self.coeffs_shape = zip(*Parallel(n_jobs=self.n_jobs,
+                                                      backend=self.backend,
+                                                      verbose=self.verbose)(
+                delayed(self._op)
+                (data[i])
+                for i in np.arange(self.n_coils)))
```
zaccharieramzi: This multi-line statement is difficult to read.
Why not:
```python
coeffs, self.coeffs_shape = zip(
	*Parallel(n_jobs=self.n_jobs, backend=self.backend, verbose=self.verbose)(
                delayed(self._op)(data[i]) for i in np.arange(self.n_coils)
	)
)
```
If too big then:
```python
coeffs, self.coeffs_shape = zip(
	*Parallel(
		n_jobs=self.n_jobs, 
		backend=self.backend, 
		verbose=self.verbose
	)(
      	delayed(self._op)(data[i]) 
		for i in np.arange(self.n_coils)
	)
)
```
or you could even define the parallel part before: 
```python
parallel_setup = Parallel(n_jobs=self.n_jobs, backend=self.backend, verbose=self.verbose)
```

mri/operators/linear/wavelet.py
```diff
@@ -57,60 +67,89 @@ def get_coeff(self):
...
             if 'array' return the data as a ndarray, otherwise return a
             pysap.Image.
-
         Returns
         -------
         data: ndarray
             the reconstructed data.
         """
-        self.transform.analysis_data = unflatten(coeffs, self.coeffs_shape)
+        self.transform.analysis_data = unflatten(coeffs, coeffs_shape)
```
zaccharieramzi: Why can't we keep it as an instance property?

mri/operators/linear/wavelet.py
```diff
@@ -57,60 +67,89 @@ def get_coeff(self):
...
+        data: ndarray
+            the reconstructed data.
+        """
+        if self.n_coils > 1:
+            images = Parallel(n_jobs=self.n_jobs,
+                              backend=self.backend,
+                              verbose=self.verbose)(
+                delayed(self._adj_op)
+                (coefs[i], self.coeffs_shape[i])
+                for i in np.arange(self.n_coils))
```
zaccharieramzi: Same remark as above for the readability of the multi-line statement

mri/operators/__init__.py
chaithyagr: Will remove them

mri/operators/linear/wavelet.py
chaithyagr: No, we use it only in calibrationless reconstruction, at least for now. Like self-calibrating reconstruction works on multichannel but, we still have one image only on which we do linear operation.

mri/operators/linear/test_wavelet_adjoint.py
chaithyagr: Yeah, will update

mri/operators/utils/__init__.py
chaithyagr: Removing

mri/operators/gradient/base.py
chaithyagr: Oops, copy issues, will fix it

mri/operators/gradient/base.py
chaithyagr: I may need help in documenting this, it is matrix and transfer matrix operation in defined function:
||Mx-y||^2
Maybe add the equation on top for reference?

mri/operators/gradient/gradient.py
chaithyagr: I think you are right. This was legacy code, will fix it.

mri/operators/gradient/gradient.py
chaithyagr: Nope, L x image shape? How do we write it?

mri/operators/gradient/gradient.py
chaithyagr: Cool

chaithyagr: although I will use vaguer terms as we implement many classes in same umbrella

mri/operators/gradient/gradient.py
chaithyagr: We have reconstructors test which I will add in next PR. sorry for this way of breakdown, will prevent this in future

mri/operators/gradient/base.py
chaithyagr: We surely need it! I personally used it in projects to be sure that the lipchtiz constant is right, especially when the user specifies it.

mri/operators/linear/wavelet.py
chaithyagr: well legacy code, and I would say it would ensure consistency in the way data is organized and also save a bit on memory (latter I am not very sure though)

mri/operators/linear/wavelet.py
chaithyagr: Will change

mri/operators/linear/wavelet.py
chaithyagr: for multichannel we need accumulated coeff. Note that we do keep it as instance property, just not here but later

mri/operators/linear/wavelet.py
zaccharieramzi: I don't understand, the number of coils is precisely the number of channels, so it means that it is for multichannel reconstruction. Calibrationless is just a way to do multichannel reconstruction.

mri/operators/gradient/base.py
zaccharieramzi: transfer matrix ? I think you mean transpose.
But yes definitely say of which function we are taking the gradient.

Also what does function handle mean?

mri/operators/gradient/gradient.py
zaccharieramzi: `[L, *data_shape]` ?

mri/operators/gradient/gradient.py
zaccharieramzi: ideally we would have an exhaustive list

mri/operators/linear/wavelet.py
chaithyagr: What I agree with is that it is used for some multichannel reconstruction. I dont agree that we use n_coils=num_channels for all multichannel reconstructions (self-calibrating reconstruction uses n_coils=1)
Given current state of codes, we use this parameter ONLY for calibrationless reconstruction and hence I named it so. 
However, I will change it given that we could have other reconstructions going forward

mri/operators/gradient/base.py
zaccharieramzi: This looks more like a unit test than a check to implement in a function.

If the user specifies it, he must be aware of what they are doing otherwise, we should have checks beforehand that check whether our provided lipschitz constants are right or not.

mri/operators/gradient/base.py
chaithyagr: Yes transpose. 
Function handle (comes from MATLAB, I am not sure what we call it in python) is basically the name of function.

eg 
```
def my_function_to_do_x
    ....
x = my_function_to_do_x
```
Here both `x` and `my_function_to_do_x` are function handles to the function "my_function_to_do_x"

mri/operators/gradient/gradient.py
chaithyagr: Done

mri/operators/gradient/base.py
chaithyagr: No, as a user I always felt it handy to launch a reconstruction that checks for lipchitz constant on its own before starting actual optimization. That way when I have result, I am very confident of lipchitz constant used.
A unittest is maybe to check if the function works right and maybe we can have it in another PR, as it is quite important. But as a user, I would not remove this.

mri/operators/linear/wavelet.py
zaccharieramzi: I don't understand how self calibrating uses only 1 coil

mri/operators/gradient/base.py
zaccharieramzi: just function then.

mri/operators/gradient/base.py
zaccharieramzi: I think this is really weird, but anyway let's not talk about it too much because this was here before, so let's leave it as is.

Just as a starter for reflexion, think about all the properties necessary for convergence that we are not testing when doing an optimisation: smoothness, agreement between gradient and function, transpose operator, ...

mri/operators/__init__.py
```diff
@@ -10,10 +10,11 @@
...
 """
 
-# Fourier operators
 from .fourier.cartesian import FFT
 from .fourier.non_cartesian import NonCartesianFFT, Stacked3DNFFT
 
-# Linear operators
 from .linear.wavelet import WaveletN, WaveletUD2
 from .linear.dictionary import DictionaryLearning
+
```
zaccharieramzi: and you can even remove the whitelines imho, but don't get stuck on that.

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,67 @@
...
+class GradBaseMRI(GradBasic):
+    """ Base Gradient class for all gradient operators
+        Implements the gradient of equation:
+        .. math::
+            ||M x - y|| ^ 2
+        Parameters
+        ----------
+        data: np.ndarray
+            input data array.
+        operator :
```
zaccharieramzi: Specify the type of `operator`.

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+# Third party import
+import numpy as np
+from modopt.math.matrix import PowerMethod
+from modopt.opt.gradient import GradBasic
+
+
+class GradBaseMRI(GradBasic):
+    """ Base Gradient class for all gradient operators
+        Implements the gradient of equation:
```
zaccharieramzi: It's not the gradient of the equation but the gradient of the function with respect to `x`.

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,67 @@
...
+        .. math::
+            ||M x - y|| ^ 2
+        Parameters
+        ----------
+        data: np.ndarray
+            input data array.
```
zaccharieramzi: specify that this is `y`.

mri/operators/gradient/base.py
zaccharieramzi: also here, it only implements `M` (not `M x`).

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class GradAnalysis(GradBaseMRI):
+    """ Gradient Analysis class.
+    This class defines the grad operators for:
+            (1/2) * sum(||Ft x - yl||^2_2,l)
+    Attributes
+    ----------
+    data: np.ndarray
+        input data array.
```
zaccharieramzi: specify that this is `y` (same for all under)

mri/operators/linear/wavelet.py
chaithyagr: (1/2) * sum(||Ft Sl Psi_t Alpha - yl||^2_2,l)

We have only one optimizing image for self calibrating reconstruction. Psi * Alpha gives one image. We apply Sl for every image and then get Fourier coefficients.

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,163 @@
...
+
+class GradSynthesis(GradBaseMRI):
+    """ Gradient Synthesis class.
+    This class defines the grad operators for:
+            (1/2) * sum(||Ft Psi_t alpha - yl||^2_2,l)
+    Attributes
+    ----------
+    data: np.ndarray
+        input 2D data array.
+    fourier_op: an object of class in mri.operators.fourier
```
zaccharieramzi: specify that this is `F` (same for all under)

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,163 @@
...
+            shape=data_shape,
+            verbose=verbose,
+            **kwargs)
+        self.fourier_op = fourier_op
+
+
+class GradSynthesis(GradBaseMRI):
+    """ Gradient Synthesis class.
+    This class defines the grad operators for:
+            (1/2) * sum(||Ft Psi_t alpha - yl||^2_2,l)
```
zaccharieramzi: Why `Ft` and not `F`?

mri/operators/gradient/base.py
chaithyagr: Hmm, I will change this in next PR? all the tests have almost passed, I would hate to have just a minor commit that would restart another set of runs for 30-45 min

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,163 @@
...
+    Attributes
+    ----------
+    data: np.ndarray
+        input 2D data array.
+    fourier_op: an object of class in mri.operators.fourier
+        a Fourier operator from FFT, NonCartesianFFT or Stacked3DNFFT
+    linear_op: an object of class in mri.operators.linear
```
zaccharieramzi: specify that this is `psi` (same for all under)

mri/operators/linear/test_wavelet_adjoint.py
```diff
@@ -107,7 +107,11 @@ def test_Wavelet_UD_2D_Multichannel(self):
         multichannel case"""
         for i in range(self.max_iter):
             print("Process Wavelet Undecimated test '{0}'...", i)
-            wavelet_op = WaveletUD2(nb_scale=4, multichannel=True, n_cpu=2)
+            wavelet_op = WaveletUD2(
+                nb_scale=4,
+                n_coils=self.num_channels,
+                n_jobs=2
```
zaccharieramzi: nitpicking here, but with a comma at the end you would even prevent unnecessary comma addition when you have a new argument to add

mri/operators/linear/wavelet.py
```diff
@@ -57,60 +67,97 @@ def get_coeff(self):
...
-        coeffs, self.coeffs_shape = flatten(self.transform.analysis_data)
+        if self.n_coils > 1:
+            coeffs, self.coeffs_shape = zip(
+                *Parallel(
+                    n_jobs=self.n_jobs,
+                    backend=self.backend,
+                    verbose=self.verbose
+                )(
+                    delayed(self._op)
+                    (data[i])
```
zaccharieramzi: do the function and the arguments have to be on separate lines?

mri/operators/gradient/gradient.py
chaithyagr: Fourier Transform, I think just that. Legacy..

mri/operators/linear/wavelet.py
chaithyagr: Nope, Well I just found it easier to read this way..

mri/operators/linear/test_wavelet_adjoint.py
chaithyagr: Hmm i didnt know that was allowed. But that would be a huge change, will just keep in mind for furture.

mri/operators/linear/wavelet.py
zaccharieramzi: so we are always only applying the wavelet transform to a single image?

chaithyagr: Except in calibrationless which optimizes for image in every channel based on proximity defined by Group LASSO or OWL.

mri/operators/gradient/gradient.py
zaccharieramzi: I think it's best to change it to `F`, it is very confusing here, since we might think it is the adjoint operator

mri/operators/gradient/base.py
```diff
@@ -0,0 +1,67 @@
...
+class GradBaseMRI(GradBasic):
+    """ Base Gradient class for all gradient operators
+        Implements the gradient of following function with respect to x:
+        .. math::
+            ||M x - y|| ^ 2
+        Parameters
+        ----------
+        data: np.ndarray
+            input data array. this is y
+        operator : a function
```
zaccharieramzi: why "a"?

mri/operators/gradient/gradient.py
```diff
@@ -0,0 +1,169 @@
...
+    This class defines the grad operators for:
+            (1/2) * sum(||Ft Sl x - yl||^2_2,l)
+    Attributes
+    ----------
+    data: np.ndarray
+        input observed data array. This is y in above equation.
+    fourier_op: an object of class in mri.operators.fourier
+        a Fourier operator from FFT, NonCartesianFFT or Stacked3DNFFT
+        This is Ft in above equation.
+    Smaps: np.ndarray
```
zaccharieramzi: Specify that this the collection of `Sl`

mri/operators/linear/wavelet.py
zaccharieramzi: I see... the name is a bit confusing then but I can't think of a better one rn. For the sake of ending this, let's leave it at that for now.

mri/operators/linear/wavelet.py
zaccharieramzi: ok this is going to be subjective so let's leave it at that

chaithyagr:
All tests have passed. The examples keep hanging at CircleCI, I have been re-starting them for now, locally I have passes. Will make changes now, @zaccharieramzi you can merge directly once you find it ok, given they are just doc changes.

zaccharieramzi:
Yes basically all the unresolved conversations are the changes I would like to have re docs

PR #55, opened 2019-11-13
Move Optimizers, update to add x_init and remove unwanted args

chaithyagr:
This is third PR in #36


mri/optimizers/forward_backward.py
```diff
@@ -0,0 +1,230 @@
...
+
+# System import
+import time
+
+# Third party import
+import numpy as np
+from modopt.opt.algorithms import ForwardBackward, POGM
+
+
+def fista(gradient_op, linear_op, prox_op, cost_op,
```
zaccharieramzi: I would be in favor of something a bit more explicit, for example `fista_reconstruction`

mri/optimizers/forward_backward.py
```diff
@@ -0,0 +1,230 @@
...
+##########################################################################
+# pySAP - Copyright (C) CEA, 2017 - 2018
+# Distributed under the terms of the CeCILL-B license, as published by
+# the CEA-CNRS-INRIA. Refer to the LICENSE file or to
+# http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.html
+# for details.
+##########################################################################
+
+"""
+FISTA or CONDAT-VU MRI reconstruction.
```
zaccharieramzi: not condat-vu

chaithyagr: Ahh POGM, will fix it

mri/optimizers/forward_backward.py
chaithyagr: No no the structure of code is such that these are just optimizers, the reconstructors are another set of classes that would call on to these functions. 
These functions just implement the optimization and the debug / intermediate result saving / plots. 
At least that is how I view these functions.
This intermediate codebase doesnt exactly make sense in that format, but the final PR fixes it.

zaccharieramzi: Hmm ok I agree

Issue #56, opened 2019-11-13
Speed up stack example and re-enable it on circleCI


chaithyagr:
We see that on circleCI the stack example being a 3D reconstruction of 128x128x160, takes too long for reconstruction and is mostly kicked by circleCI causing crashes and failures.
for now in #36, through #55  I plan to disable the running of this example on CircleCI to ease the flow of code refactoring. 
I am ensuring that the example still runs fine and I will speed it up and re-enable it on CI once we have stable mainline.

chaithyagr:
This is now fixed and speeded up by taking only a part of kspace. This is merged at mainline and reenabled on circle ci, closing..

PR #58, opened 2019-11-18
Implements the undersampling of kspace z direction

mathrip:
This PR implements the undersampling of the k-space in the z-direction (only working for the stacks implementations).

This solves #56 as the z-dimension has been cropped, it also solves #33

### DIFF
--- a/examples/non_cartesian_reconstruction_stack_3d.py
+++ b/examples/non_cartesian_reconstruction_stack_3d.py
@@ -100,7 +100,7 @@ print('The Base SSIM is : ' + str(base_ssim))
 
 # TODO get the right mu operator
 # Setup the operators
-linear_op = WaveletN(wavelet_name="sym8", nb_scales=4)
+linear_op = WaveletN(wavelet_name="sym8", nb_scales=4, dim=3)
 regularizer_op = SparseThreshold(Identity(), 6 * 1e-9, thresh_type="soft")
 # Setup Reconstructor
 reconstructor = SingleChannelReconstructor(
--- a/examples/non_cartesian_reconstruction_stack_3d.py
+++ b/examples/non_cartesian_reconstruction_stack_3d.py
@@ -16,7 +16,8 @@ We use the toy datasets available in pysap, more specifically a 3D Orange.
 # Package import
 from mri.operators import Stacked3DNFFT, WaveletN
 from mri.operators.utils import convert_locations_to_mask, \
-    gridded_inverse_fourier_transform_stack, get_stacks_fourier
+    gridded_inverse_fourier_transform_stack, get_stacks_fourier, \
+    convert_mask_to_locations
 from mri.reconstructors import SingleChannelReconstructor
 import pysap
 from pysap.data import get_sample_data
@@ -31,14 +32,25 @@ import numpy as np
 image = get_sample_data('3d-pmri')
 image = pysap.Image(data=np.sqrt(np.sum(np.abs(image.data)**2, axis=0)))
 
-# Obtain MRI non-cartesian mask
-radial_mask = get_sample_data("mri-radial-samples")
-z_locations = np.repeat(np.linspace(-0.5, 0.5, image.shape[2], endpoint=False),
-                        radial_mask.shape[0])
+# Reducing the size of the volume for faster computation
+image.data = image.data[:, :, 48: -48]
+
+# Obtain MRI non-cartesian sampling plane
+mask_radial = get_sample_data("mri-radial-samples")
+
+# Tiling the plane on the z-direction
+# sampling_z = np.ones(image.shape[2])  # no sampling
+sampling_z = np.random.randint(2, size=image.shape[2])  # random sampling
+sampling_z[22: 42] = 1
+Nz = sampling_z.sum()  # Number of acquired plane
+
+z_locations = np.repeat(convert_mask_to_locations(sampling_z),
+                        mask_radial.shape[0])
 z_locations = z_locations[:, np.newaxis]
-kspace_loc = np.hstack([np.tile(radial_mask.data, (image.shape[2], 1)),
+kspace_loc = np.hstack([np.tile(mask_radial.data, (Nz, 1)),
                         z_locations])
-mask = pysap.Image(data=convert_locations_to_mask(kspace_loc, image.shape))
+mask = pysap.Image(data=np.moveaxis(
+    convert_locations_to_mask(kspace_loc, image.shape), -1, 0))
 
 # View Input
 # image.show()
@@ -60,15 +72,20 @@ fourier_op = Stacked3DNFFT(kspace_loc=kspace_loc,
 kspace_obs = fourier_op.op(image.data)
 
 # Gridded solution
-grid_space = [np.linspace(-0.5, 0.5, num=image.shape[i])
-              for i in range(len(image.shape) - 1)]
+grid_space = [np.linspace(-0.5, 0.5, num=img_shape)
+              for img_shape in image.shape[:-1]]
 grid = np.meshgrid(*tuple(grid_space))
-kspace_plane_loc, z_sample_loc, sort_pos = get_stacks_fourier(kspace_loc)
-grid_soln = gridded_inverse_fourier_transform_stack(kspace_plane_loc,
-                                                    z_sample_loc,
-                                                    kspace_obs,
-                                                    tuple(grid),
-                                                    'linear')
+kspace_plane_loc, z_sample_loc, sort_pos, idx_mask_z = get_stacks_fourier(
+    kspace_loc,
+    image.shape)
+grid_soln = gridded_inverse_fourier_transform_stack(
+    kspace_data_sorted=kspace_obs[sort_pos],
+    kspace_plane_loc=kspace_plane_loc,
+    idx_mask_z=idx_mask_z,
+    grid=tuple(grid),
+    volume_shape=image.shape,
+    method='linear')
+
 image_rec0 = pysap.Image(data=grid_soln)
 # image_rec0.show()
 base_ssim = ssim(image_rec0, image)
--- a/mri/operators/fourier/non_cartesian.py
+++ b/mri/operators/fourier/non_cartesian.py
@@ -482,8 +482,12 @@ class Stacked3DNFFT(OperatorBase):
         """
         self.num_slices = shape[2]
         self.shape = shape
-        (kspace_plane_loc, self.z_sample_loc,
-         self.sort_pos) = get_stacks_fourier(kspace_loc)
+        (kspace_plane_loc, self.z_sample_loc, self.sort_pos, self.idx_mask_z) \
+            = \
+            get_stacks_fourier(
+            kspace_loc,
+            self.shape,
+            )
         self.acq_num_slices = len(self.z_sample_loc)
         self.stack_len = len(kspace_plane_loc)
         self.plane_fourier_operator = \
@@ -494,18 +498,20 @@ class Stacked3DNFFT(OperatorBase):
     def _op(self, data):
         fft_along_z_axis = np.fft.fftshift(np.fft.fft(
             np.fft.ifftshift(data, axes=2),
-            n=self.acq_num_slices,
             norm="ortho"),
             axes=2)
-        stacked_kspace = np.asarray(
-            [self.plane_fourier_operator.op(fft_along_z_axis[:, :, slice])
-             for slice in np.arange(self.acq_num_slices)])
-        stacked_kspace = np.reshape(stacked_kspace,
-                                    self.acq_num_slices * self.stack_len)
+        stacked_kspace_sampled = np.asarray(
+            [self.plane_fourier_operator.op(fft_along_z_axis[:, :, stack])
+             for stack in self.idx_mask_z])
+        stacked_kspace_sampled = np.reshape(
+            stacked_kspace_sampled,
+            self.acq_num_slices * self.stack_len)
         # Unsort the Coefficients
         inv_idx = np.zeros_like(self.sort_pos)
         inv_idx[self.sort_pos] = np.arange(len(self.sort_pos))
-        return stacked_kspace[inv_idx]
+        # Return kspace unsorted and normalised by the ratio of slices acquired
+        return stacked_kspace_sampled[inv_idx] * \
+            np.sqrt(self.num_slices / self.acq_num_slices)
 
     def op(self, data):
         """ This method calculates Fourier transform.
@@ -531,24 +537,20 @@ class Stacked3DNFFT(OperatorBase):
     def _adj_op(self, coeff):
         coeff = coeff[self.sort_pos]
         stacks = np.reshape(coeff, (self.acq_num_slices, self.stack_len))
-        # Receive First Fourier transformed data (per plane) in Nz x N x N
-        adj_fft_along_z_axis = \
-            np.asarray([self.plane_fourier_operator.adj_op(stacks[slice]).T
-                        for slice in np.arange(stacks.shape[0])])
-        # TODO fix for higher N, this is not a usecase
-        # interpolate_kspace = interp1d(self.z_samples[:, 0],
-        #                               adj_fft_along_z_axis, kind='zero',
-        #                               axis=0, bounds_error=False,
-        #                               fill_value=0)
-        # adj_fft_along_z_axis = interpolate_kspace(np.linspace(-0.5, 0.5,
-        #                                            self.num_slices,
-        #                                            endpoint=False))
+        # Receive First Fourier transformed data (per plane) in N x N x Nz
+        adj_fft_along_z_axis = np.zeros((*self.plane_fourier_operator.shape,
+                                         self.num_slices),
+                                        dtype=coeff.dtype)
+        for idxs, idxm in enumerate(self.idx_mask_z):
+            adj_fft_along_z_axis[:, :, idxm] = \
+                self.plane_fourier_operator.adj_op(stacks[idxs])
+
         stacked_images = np.fft.ifftshift(np.fft.ifft(
-            np.asarray(np.fft.fftshift(adj_fft_along_z_axis, axes=0)),
-            axis=0, n=self.num_slices, norm="ortho"),
-            axes=0)
-        # Transpose the data to place in form of Nx x Ny x Nz
-        return stacked_images.T
+                np.asarray(np.fft.fftshift(adj_fft_along_z_axis, axes=-1)),
+                axis=-1, norm="ortho"),
+            axes=-1)
+
+        return stacked_images * np.sqrt(self.num_slices / self.acq_num_slices)
 
     def adj_op(self, coeff):
         """ This method calculates inverse masked non-uniform Fourier
--- a/mri/operators/fourier/utils.py
+++ b/mri/operators/fourier/utils.py
@@ -110,33 +110,51 @@ def normalize_frequency_locations(samples, Kmax=None):
     return samples_locations
 
 
-def get_stacks_fourier(kspace_loc):
+def get_stacks_fourier(kspace_loc, volume_shape):
     """Function that splits an incoming 3D stacked k-space samples
     into a 2D non-Cartesian plane and the vector containing the z k-space
-    values of all the plane and converts to stacks of 2D. This function also
-    checks for any issues of the incoming k-space pattern and if the stack
-    property is not satisfied.
+    values of the stacks acquiered and converts to stacks of 2D.
+    This function also checks for any issues of the incoming k-space
+    pattern and if the stack property is not satisfied.
     Stack Property: The k-space locations originate from a stack of 2D samples.
 
     Parameters
     ----------
-    ksapce_plane_loc: np.ndarray
-        the mask samples in the 3D Fourier domain.
-
+    kspace_loc: np.ndarray
+        Acquired 3D k-space locations : stacks of same non-Cartesian samples,
+        while Cartesian under-sampling on the stacks direction.
+    volume_shape: tuple
+        Reconstructed volume shape
     Returns
     ----------
-    ksapce_plane_loc: np.ndarray
+    kspace_plane_loc: np.ndarray
         A 2D array of samples which when stacked gives the 3D samples
     z_sample_loc: np.ndarray
         A 1D array of z-sample locations
     sort_pos: np.ndarray
         The sorting positions for opertor and inverse for incoming data
+    idx_mask_z: np.ndarray
+        contains the indices of the acquired Fourier planes (z direction)
     """
     # Sort the incoming data based on Z, Y then X coordinates
     # This is done for easier stacking
     sort_pos = np.lexsort(tuple(kspace_loc[:, i]
                                 for i in np.arange(3)))
     kspace_loc = kspace_loc[sort_pos]
+
+    # Find the mask used to sample stacks in z direction
+    full_stack_z_loc = convert_mask_to_locations(
+        np.ones(volume_shape[2]),
+    )[:, 0]
+    sampled_stack_z_loc = np.unique(kspace_loc[:, 2])
+
+    try:
+        idx_mask_z = np.asarray([
+            np.where(x == full_stack_z_loc)[0][0] for x in sampled_stack_z_loc
+        ])
+    except IndexError:
+        raise ValueError('The input must be a stack of 2D k-Space data')
+
     first_stack_len = np.size(np.where(
         kspace_loc[:, 2] == np.min(kspace_loc[:, 2])))
     acq_num_slices = int(len(kspace_loc) / first_stack_len)
@@ -150,10 +168,10 @@ def get_stacks_fourier(kspace_loc):
             or not np.all(stacked[:, :, 0:2] == stacked[0, :, 0:2]) \
             or not np.all(stacked[:, :, 2] == z_expected_stacked):
         raise ValueError('The input must be a stack of 2D k-Space data')
-    ksapce_plane_loc = stacked[0, :, 0:2]
+    kspace_plane_loc = stacked[0, :, 0:2]
     z_sample_loc = stacked[:, 0, 2]
     z_sample_loc = z_sample_loc[:, np.newaxis]
-    return ksapce_plane_loc, z_sample_loc, sort_pos
+    return kspace_plane_loc, z_sample_loc, sort_pos, idx_mask_z
 
 
 def gridded_inverse_fourier_transform_nd(kspace_loc,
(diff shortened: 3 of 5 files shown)


mri/operators/fourier/utils.py
```diff
@@ -257,7 +257,8 @@ def get_stacks_fourier(kspace_loc):
     ----------
     ksapce_plane_loc: np.ndarray
```
philouc: typo: ksapce -> kspace

mri/operators/fourier/utils.py
```diff
@@ -257,7 +257,8 @@ def get_stacks_fourier(kspace_loc):
...
         the mask samples in the 3D Fourier domain.
-
+    shape: tuple
+        Reconstructed volume shape
     Returns
     ----------
     ksapce_plane_loc: np.ndarray
```
philouc: same typo

mri/operators/fourier/utils.py
```diff
@@ -266,15 +267,26 @@ def get_stacks_fourier(kspace_loc):
         A 1D array of z-sample locations
     sort_pos: np.ndarray
         The sorting positions for opertor and inverse for incoming data
+    idx_mask_z: np.ndarray
+        contains the indices of the acquired Fourier plane
```
philouc: Fourier planes instead of plane?

mri/operators/fourier/utils.py
philouc: update name variable given the typo

mri/operators/fourier/utils.py
```diff
@@ -288,4 +300,4 @@ def get_stacks_fourier(kspace_loc):
     ksapce_plane_loc = stacked[0, :, 0:2]
     z_sample_loc = stacked[:, 0, 2]
     z_sample_loc = z_sample_loc[:, np.newaxis]
-    return ksapce_plane_loc, z_sample_loc, sort_pos
+    return ksapce_plane_loc, z_sample_loc, sort_pos, idx_mask_z
```
philouc: propagate changes on name variable

mri/operators/fourier/utils.py
```diff
@@ -255,37 +255,53 @@ def get_stacks_fourier(kspace_loc):
...
     kspace_loc = kspace_loc[sort_pos]
-    first_stack_len = np.size(np.where(
-        kspace_loc[:, 2] == np.min(kspace_loc[:, 2])))
+
+    # Find the mask used to sample stacks in z direction
+    full_stack_loc = convert_mask_to_locations(np.ones(shape[2]))[:, 0]
+    sampled_stack_loc = np.unique(kspace_loc[:, 2])
+
+    try:
+        idx_mask_z = np.asarray([np.where(x == full_stack_loc)[0][0] for
```
zaccharieramzi: It feels like this doesn't really belong to the function. It's doing something completely parallel to the rest of it. Why does it need to be in this function ?

mri/operators/fourier/utils.py
```diff
@@ -255,37 +255,53 @@ def get_stacks_fourier(kspace_loc):
...
+                                       np.min(kspace_loc[:, 2])))
     acq_num_slices = int(len(kspace_loc) / first_stack_len)
+
     stacked = np.reshape(kspace_loc, (acq_num_slices,
                                       first_stack_len, 3))
     z_expected_stacked = np.reshape(np.repeat(stacked[:, 0, 2],
                                               first_stack_len),
                                     (acq_num_slices,
                                      first_stack_len))
+
```
zaccharieramzi: Let's try to avoid as much as possible whiteline changes. They mess up the git history and clutter the PR.

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -94,25 +94,30 @@ def gridded_inverse_fourier_transform_nd(kspace_loc,
...
-        This holds the z-sample locations for stacks. Again, extracted using
+    idx_mask_z: np.ndarray
+        contains the indices of the acquired Fourier plane. Extracted using
         get_stacks_fourier function
-    kspace_data: np.ndarray
-        The k-space data corresponding to kspace_plane_loc above
-    grid: np.ndarray
+    grid: tuple
         The Gridded matrix for which you want to calculate k_space Smaps
+    shape: tuple
```
zaccharieramzi: same, maybe `volume_shape` is more adapted.

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -121,17 +126,16 @@ def gridded_inverse_fourier_transform_stack(kspace_plane_loc, z_sample_loc,
...
-                     fill_value=0))
-    # Move the slice axis to last : Make to Nx x Ny x Nz
-    gridded_kspace = np.moveaxis(np.asarray(gridded_kspace), 0, 2)
+    for i, idx_z in enumerate(idx_mask_z):
+        gridded_kspace[:, :, idx_z] = griddata(
+            kspace_plane_loc,
+            kspace_data_sorted[i*stack_len:(i+1)*stack_len],
+            grid,
+            method=method,
+            fill_value=0)
```
zaccharieramzi: nitpicking here, but to handle multi-line in the best way possible you can even do:
```python
gridded_kspace[:, :, idx_z] = griddata(
  kspace_plane_loc,
  kspace_data_sorted[i*stack_len:(i+1)*stack_len],
  grid,  
  method=method,
  fill_value=0,
)
```
If for example someone wants to add an argument at the end, they don't need to modify the `fill_value` line too.

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -215,8 +219,8 @@ def get_Smaps(k_space, img_shape, samples, thresh,
                                   endpoint=False)
                       for i in np.arange(np.size(img_shape)-1)]
         grid = np.meshgrid(*grid_space)
-        kspace_plane_loc, z_sample_loc, sort_pos = \
-            get_stacks_fourier(samples)
+        kspace_plane_loc, _, sort_pos, idx_mask_z = \
```
zaccharieramzi: Then do we need to have `z_sample_loc` returned in the `get_stacks_fourier` function?

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -94,25 +94,30 @@ def gridded_inverse_fourier_transform_nd(kspace_loc,
...
         in plane, extracted using get_stacks_fourier function
-    z_sample_loc: np.ndarray
-        This holds the z-sample locations for stacks. Again, extracted using
+    idx_mask_z: np.ndarray
+        contains the indices of the acquired Fourier plane. Extracted using
         get_stacks_fourier function
-    kspace_data: np.ndarray
-        The k-space data corresponding to kspace_plane_loc above
-    grid: np.ndarray
+    grid: tuple
```
zaccharieramzi: why did the type change?

mri/operators/fourier/non_cartesian.py
```diff
@@ -482,8 +482,12 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
         """
         self.num_slices = shape[2]
         self.shape = shape
-        (kspace_plane_loc, self.z_sample_loc,
-         self.sort_pos) = get_stacks_fourier(kspace_loc)
+        (kspace_plane_loc,
+         self.z_sample_loc,
+         self.sort_pos,
+         self.idx_mask_z) = get_stacks_fourier(
```
zaccharieramzi: I think this can be on a single line (it's just 91 characters)

mri/operators/fourier/non_cartesian.py
```diff
@@ -494,18 +498,25 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
     def _op(self, data):
         fft_along_z_axis = np.fft.fftshift(np.fft.fft(
             np.fft.ifftshift(data, axes=2),
-            n=self.acq_num_slices,
```
zaccharieramzi: why was this removed?

mri/operators/fourier/non_cartesian.py
```diff
@@ -494,18 +498,25 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
...
             norm="ortho"),
             axes=2)
+
         stacked_kspace = np.asarray(
             [self.plane_fourier_operator.op(fft_along_z_axis[:, :, slice])
-             for slice in np.arange(self.acq_num_slices)])
-        stacked_kspace = np.reshape(stacked_kspace,
-                                    self.acq_num_slices * self.stack_len)
+             for slice in range(self.num_slices)])
```
zaccharieramzi: while we are changing this line, let's use a different variable name, `slice` is protected in Python.

mri/operators/fourier/non_cartesian.py
```diff
@@ -494,18 +498,25 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
...
+        stacked_kspace_sampled = np.zeros((self.acq_num_slices,
```
zaccharieramzi: I think this is legacy

mri/operators/fourier/non_cartesian.py
```diff
@@ -494,18 +498,25 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
...
+                                           self.stack_len),
+                                          dtype=data.dtype)
+        stacked_kspace_sampled = stacked_kspace[self.idx_mask_z, :]
```
zaccharieramzi: You can just do `stacked_kspace_sampled = stacked_kspace[self.idx_mask_z]`

mri/operators/fourier/non_cartesian.py
```diff
@@ -494,18 +498,25 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
...
+
+        stacked_kspace_sampled = np.reshape(
+            stacked_kspace_sampled,
+            self.acq_num_slices * self.stack_len)
         # Unsort the Coefficients
         inv_idx = np.zeros_like(self.sort_pos)
         inv_idx[self.sort_pos] = np.arange(len(self.sort_pos))
-        return stacked_kspace[inv_idx]
+        return stacked_kspace_sampled[inv_idx] * \
+            np.sqrt(self.num_slices / self.acq_num_slices)
```
zaccharieramzi: Can you add a comment on this normalisation?

mri/operators/fourier/test_fourier_adjoint.py
```diff
@@ -196,8 +196,14 @@ def test_similarity_stack_3D(self):
...
                 _mask = np.random.randint(2, size=(N, N))
-                _mask3D = np.asarray([_mask for i in np.arange(Nz)])
-                _samples = convert_mask_to_locations(_mask3D.swapaxes(0, 2))
+
+                # Generate random mask along z
+                sampling_z = np.random.randint(2, size=Nz)
+                _mask3D = np.zeros((N, N, Nz))
+                for idx, acq_z in enumerate(sampling_z):
+                    _mask3D[:, :, idx] = _mask * acq_z
+                _samples = convert_mask_to_locations(_mask3D)
```
zaccharieramzi: I can see that they were here before, but I don't understand why these variable names begin with `_`.

mri/operators/fourier/utils.py
mathrip: This function returns the characteristics of the k-space, the mask of stacks in z-direction is one of its characteristics. I think this need to be returned by this function.

mri/parallel_mri/extract_sensitivity_maps.py
mathrip: idx_mask_z is needed. z_sample_loc is redondant as we already have the kspace_data to give the z-locations.

mri/parallel_mri/extract_sensitivity_maps.py
mathrip: z_sample_loc is requiered in other call of the function get_stacks_fourier

mri/operators/fourier/utils.py
```diff
@@ -245,7 +245,7 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
     return gradient_op, linear_op, prox_op, cost_op
 
 
-def get_stacks_fourier(kspace_loc):
+def get_stacks_fourier(kspace_loc, shape):
```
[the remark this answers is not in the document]
mathrip: I agree, however, there is a lot of dependencies for this function, and a change of variable name would lead to change multiples calls in the examples and test scripts. It is best if it stays as "shape" for now.

mri/operators/fourier/utils.py
```diff
@@ -255,37 +255,53 @@ def get_stacks_fourier(kspace_loc):
...
     Returns
     ----------
-    ksapce_plane_loc: np.ndarray
+    kspace_plane_loc: np.ndarray
         A 2D array of samples which when stacked gives the 3D samples
     z_sample_loc: np.ndarray
         A 1D array of z-sample locations
     sort_pos: np.ndarray
         The sorting positions for opertor and inverse for incoming data
+    idx_mask_z: np.ndarray
```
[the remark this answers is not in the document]
LElgueddari: What do you mean by masking entire planes?
Masking actually means not acquiring so the doc should mention that it's possible to not acquire entire plane, is that what you mean?

mri/operators/fourier/utils.py
```diff
@@ -248,15 +248,16 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
...
-    checks for any issues of the incoming k-space pattern and if the stack
-    property is not satisfied.
+    values of the stacks acquiered and converts to stacks of 2D.
+    This function also checks for any issues of the incoming k-space
+    pattern and if the stack property is not satisfied.
     Stack Property: The k-space locations originate from a stack of 2D samples.
 
     Parameters
     ----------
     kspace_plane_loc: np.ndarray
```
LElgueddari: The parameter name is `kspace_loc` not `kspace_plane_loc` which actually corresponds to the acquired k-space samples position in the 3 dimension.

mri/operators/fourier/utils.py
```diff
@@ -248,15 +248,16 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
...
-        the mask samples in the 3D Fourier domain.
+        the mask samples in the 3D Fourier domain
+        possibility to mask entire planes (sampling in z direction)
```
LElgueddari: I'm sorry but this doesn't make any sense to me

mri/parallel_mri/extract_sensitivity_maps.py
```diff
@@ -134,7 +134,8 @@ def gridded_inverse_fourier_transform_stack(kspace_data_sorted,
             kspace_data_sorted[i*stack_len:(i+1)*stack_len],
             grid,
             method=method,
-            fill_value=0)
+            fill_value=0
```
LElgueddari: I think that what @zaccharieramzi  asked is to add a comma after `fill_value=0`, otherwise there is no point of doing this

examples/non_cartesian_reconstruction_stack_3d.py
```diff
@@ -60,15 +74,20 @@
...
-kspace_plane_loc, z_sample_loc, sort_pos = get_stacks_fourier(kspace_loc)
-grid_soln = gridded_inverse_fourier_transform_stack(kspace_plane_loc,
-                                                    z_sample_loc,
-                                                    kspace_obs,
-                                                    tuple(grid),
-                                                    'linear')
+kspace_plane_loc, z_sample_loc, sort_pos, idx_mask_z = get_stacks_fourier(
+    kspace_loc,
+    image.shape)
+grid_soln = gridded_inverse_fourier_transform_stack(
```
chaithyagr: Ahh, this is bad, we would want the stacking done by the function itself. Can we have an issue for this?
We can rename this function and write another function that actually does gridded_inverse_fourier transform that takes in just the stack.

mri/operators/fourier/non_cartesian.py
chaithyagr: Would this be better?
```python
(kspace_plane_loc, self.z_sample_loc, self.sort_pos, self.idx_mask_z) \
            = \
            get_stacks_fourier(
                kspace_loc,
                self.shape,
            )
```

I think @zaccharieramzi  is best to suggest the right way here, but to me the above is more readable.

mri/operators/fourier/non_cartesian.py
chaithyagr: Yes, we can remove this based on new codes, this seems like a redundant line here

mri/operators/fourier/non_cartesian.py
chaithyagr: @zaccharieramzi this is removed to have a full FFT, they will use this to later choose what undersampled data they want.

mri/operators/fourier/non_cartesian.py
```diff
@@ -494,18 +498,25 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
...
-            [self.plane_fourier_operator.op(fft_along_z_axis[:, :, slice])
-             for slice in np.arange(self.acq_num_slices)])
-        stacked_kspace = np.reshape(stacked_kspace,
-                                    self.acq_num_slices * self.stack_len)
+            [self.plane_fourier_operator.op(fft_along_z_axis[:, :, stack])
+             for stack in range(self.num_slices)])
+        stacked_kspace_sampled = np.zeros((self.acq_num_slices,
+                                           self.stack_len),
+                                          dtype=data.dtype)
+        stacked_kspace_sampled = stacked_kspace[self.idx_mask_z]
```
chaithyagr: Please correct me if my understanding here is wrong:
You are effectively doing a complete 2D NFFT on each slice location and choosing undersampled data.
Can we move to just do NFFT on only the slices where we want?

Something like:

```python
stacked_kspace_sampled = np.asarray(
            [self.plane_fourier_operator.op(fft_along_z_axis[:, :, stack])
             for stack in self.idx_mask_z])
```

Was this explored and would this cause any issues?

mri/operators/fourier/non_cartesian.py
```diff
@@ -526,29 +537,26 @@ def op(self, data):
             coeff = [self._op(data[i])
                      for i in range(self.n_coils)]
         coeff = np.asarray(coeff)
+
         return coeff
 
     def _adj_op(self, coeff):
         coeff = coeff[self.sort_pos]
         stacks = np.reshape(coeff, (self.acq_num_slices, self.stack_len))
         # Receive First Fourier transformed data (per plane) in Nz x N x N
```
chaithyagr: It looks like based on code update, you are organizing data as `N x N x Nz`. Can you please update this comment?
Also, it would help to mention this in doc if not done already.

mri/operators/fourier/utils.py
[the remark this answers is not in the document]
chaithyagr: You can use refactor in pycharm for this, its good practice to use the right names from start, can we please get this?

mri/parallel_mri/extract_sensitivity_maps.py
chaithyagr: They both represent the same thing after all right? One is z_locations and other just indexes to these locations, can we open an issue to clean this function up? It would be great to get in this PR also, but that's your call..

examples/non_cartesian_reconstruction_stack_3d.py
mathrip: done !

mathrip: done

mri/operators/fourier/utils.py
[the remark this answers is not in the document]
mathrip: I know, the problem is that it will affect some functions and scripts not related to this PR. I think this is a typo problem that needs to be fixed in a special PR to improve the code, make it user-friendly. I raised an issue #61

mri/operators/fourier/non_cartesian.py
mathrip: done !

mri/operators/fourier/utils.py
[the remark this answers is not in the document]
chaithyagr: I am sorry but I dont understand. We are not referring to all the refactoring mentioned in #61 . This PR adds new parameter shape. Now we would prefer if the code enters with right nomenclature directly. Other name changes can be done in another PR. But this particular one looks to me like is added in this PR and it only makes sense to add the right code in first commit than fix it up later.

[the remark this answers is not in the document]
chaithyagr: I am sorry but I dont understand. We are not referring to all the refactoring mentioned in #61 . This PR adds new parameter shape. Now we would prefer if the code enters with right nomenclature directly. Other name changes can be done in another PR. But this particular one looks to me like is added in this PR and it only makes sense to add the right code in first commit than fix it up later.

mri/operators/fourier/non_cartesian.py
mathrip: I made the change, thank you

mri/operators/fourier/non_cartesian.py
mathrip: Comment updated. There is already mention of the shape N x N x Nz in the function doc, I don't touch it.

mri/operators/fourier/utils.py
mathrip: whiteline removed

mri/parallel_mri/extract_sensitivity_maps.py
mathrip: I raised an issue and will try to work on that for this PR

mri/operators/fourier/utils.py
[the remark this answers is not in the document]
zaccharieramzi: Yes, from what I understand now we have the possibility to undersample in the stack direction as well right?

mri/operators/fourier/utils.py
zaccharieramzi: I don't see it removed

mri/parallel_mri/extract_sensitivity_maps.py
zaccharieramzi: No I'm asking why the type of `grid` changed

mri/operators/fourier/utils.py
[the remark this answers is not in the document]
LElgueddari: Yes we do. It is just the term mask that do not make that much sense. We do not really mask the 3D k-space since each plane is non-Cartesian therefore it cannot be considered as masked Fourier space. 
Beacause of this specific point the following documentation is false
https://github.com/CEA-COSMIC/pysap-mri/blob/cf4dff5dcb28058d946fbf210e584ededa727ec8/mri/operators/fourier/utils.py#L258-L259
It should be replaced by 
```
Acquired 3D k-space locations such that the each plane is composed of the same non-Cartesian samples, 
while the third dimension correspond to a Cartesian under-sampling
```
Something like this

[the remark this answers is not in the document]
zaccharieramzi: I see !

mri/parallel_mri/extract_sensitivity_maps.py
mathrip: Because we realised it is a tuple (containing two arrays). 
Do you prefere if I precise tuple of np.array?

mri/operators/fourier/utils.py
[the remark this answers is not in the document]
zaccharieramzi: @mathrip, I am not sure I understand your point. You introduced this variable so it means that everywhere the keyword argument `shape` is used for this function is because you introduced it, or am I mistaken?

mri/operators/fourier/utils.py
```diff
@@ -245,33 +245,50 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
...
+        contains the indices of the acquired Fourier planes (z direction)
     """
     # Sort the incoming data based on Z, Y then X coordinates
     # This is done for easier stacking
     sort_pos = np.lexsort(tuple(kspace_loc[:, i]
                                 for i in np.arange(3)))
     kspace_loc = kspace_loc[sort_pos]
+
+    # Find the mask used to sample stacks in z direction
+    full_stack_z_loc = \
```
zaccharieramzi: I think it's preferable to limit the use of `\`, you could do:
```python
full_stack_z_loc = convert_mask_to_locations(
  np.ones(volume_shape[2]),
)[:, 0]

File: mri/operators/fourier/utils.py [hunk 27]
```diff
@@ -245,33 +245,50 @@ def generate_operators(data, wavelet_name, samples, mu=1e-06, nb_scales=4,
...
+        convert_mask_to_locations(np.ones(volume_shape[2]))[:, 0]
+    sampled_stack_z_loc = np.unique(kspace_loc[:, 2])
+
+    try:
```
zaccharieramzi: I think it would be nice to include a comment saying what exactly the test is here.
I understood the first time, and already forgot it...

mri/parallel_mri/extract_sensitivity_maps.py
zaccharieramzi: ok the grid is expressed as a tuple of `ndarray`s?
Maybe not in the type, but in the description of the variable it would be nice.

mri/operators/fourier/non_cartesian.py
```diff
@@ -482,8 +482,12 @@ def __init__(self, kspace_loc, shape, implementation='cpu', n_coils=1):
         """
         self.num_slices = shape[2]
         self.shape = shape
-        (kspace_plane_loc, self.z_sample_loc,
-         self.sort_pos) = get_stacks_fourier(kspace_loc)
+        (kspace_plane_loc, self.z_sample_loc, self.sort_pos, self.idx_mask_z) \
```
zaccharieramzi: why can't we make this:
```python
(kspace_plane_loc, self.z_sample_loc, self.sort_pos, self.idx_mask_z) = get_stacks_fourier(
  kspace_loc,
  self.shape,
)
```
?

chaithyagr: That would cross the PEP8 line length of 78 characters right?

zaccharieramzi: https://www.python.org/dev/peps/pep-0008/#maximum-line-length
Since we are the only ones working on this project, I think it's fine to have a longer maximum line length like 99.

In the team I worked in before, because it was not an open-source project, we even used 120 characters.

mri/parallel_mri/extract_sensitivity_maps.py
mathrip: Ok!

mri/operators/fourier/utils.py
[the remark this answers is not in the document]
mathrip: Yes you are right. So I changed it in this function because I had this argument. But it exists also in other functions such as extract_sensitivity, and the class Stacked3DNFFT. I thought you wanted me to change the name here as well... but I was mistaken !

mri/operators/fourier/utils.py
mathrip: Ok, it will be on the next commit

mri/operators/fourier/utils.py
mathrip: done

mri/operators/fourier/non_cartesian.py
mathrip: It failed the test if I change it, because of pep8. I let it like this for now, but if you insist I'll change it.

mri/operators/fourier/utils.py
mathrip: Ok done

mri/operators/fourier/non_cartesian.py
```diff
@@ -526,29 +537,26 @@ def op(self, data):
             coeff = [self._op(data[i])
                      for i in range(self.n_coils)]
         coeff = np.asarray(coeff)
+
```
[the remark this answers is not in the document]
mathrip: Ok, it's done

chaithyagr:
We may have to rebase this completely. This could be a hard task, based on changes we can as well take a new branch and do it manually. @mathrip do you want to do this? Or shall I take over?

chaithyagr:
Taking over for rebasing and ensuring this is at mainline

chaithyagr:
I have rebased all the codes to master. @zaccharieramzi can you please do a final check? 
I will merge once I have tests passing

zaccharieramzi:
@mathrip have you had the chance to re-read the rebased version?

mathrip:
> @mathrip have you had the chance to re-read the rebased version?

Not yet, but I'll do it tomorrow !

mathrip:
> @mathrip have you had the chance to re-read the rebased version?

I reviewed and I approve @zaccharieramzi can you approve it? Thanks

PR #63, opened 2020-01-06
Fix for Travis

chaithyagr:
This PR holds:
1) Fixes for Travis to enable pynfft import to work right: https://github.com/scikit-learn/scikit-learn/issues/14485#issuecomment-522218682
2) Limit Scipy to 1.3.3 as 1.4 is breaking some things for wavelet UD tests for now. This must be fixed in Modopt as we leverage functions from Modopt. This fix here is a temporary workaround

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: python
 
 matrix:
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ Jean-Luc Starck <jl.stark@cea.fr>
 Philippe Ciuciu <philippe.ciuciu@cea.fr>
 """
 # Write setup
-setup_requires = ["numpy", "scipy", "cython", "pytest-runner", "scikit-image"]
+setup_requires = ["numpy", "cython", "pytest-runner", "scikit-image"]
 
 pip_main(['install'] + setup_requires)
 
@@ -49,6 +49,7 @@ setup(
     install_requires=["scikit-learn",
                       "progressbar2",
                       "joblib",
+                      "scipy<=1.3.3",
                       "psutil",
                       "pynfft@git+https://github.com/ghisvail/pyNFFT@master"],
     dependency_links=['https://github.com/ghisvail/pyNFFT/tarball/master#egg=pynfft-1.3.0'],


zaccharieramzi:
I'd rather have @sfarrens or @AGrigis 's opinion on whether we can get rid of the `trusty` distribution. As I mentionned on slack, I don't know whether this is important for some reason. Also, it might mean not supporting some Ubuntu version which might need to be indicated somewhere.

Finally, you should definitely open an issue related to the `scipy` version problem.

Otherwise, lgtm if all green.

chaithyagr:
Well, FWIW, looks like we have many other issues to be cleared to get Travis up. I would really like to use this PR to fix everything. At this point I am adding WIP as a tag to PR, Please ignore any updates here till I remove that tag

chaithyagr:
We have passed Travis again finally!!! Yay.
Now the circleci is broken due to https://github.com/CEA-COSMIC/pysap/issues/124, and I have a PR being prepared for it.
For now we can merge this if it looks fine @zaccharieramzi

zaccharieramzi:
As said on slack, I am fine with this hot-fix, but it should be considered one.
We need to document why it doesn't work when removing `trusty` (i.e. ubuntu 14.04) and going to `xenial` (i.e. ubuntu 16.04). And the corresponding issue cannot be closed.

chaithyagr:
So shall we merge for now to keep travis up? The circle ci PR is in flight https://github.com/CEA-COSMIC/pysap/pull/125

.travis.yml
```diff
@@ -1,6 +1,6 @@
 sudo: required
-dist: trusty
 language: python
+dist: trusty
```
zaccharieramzi: let's have no difference here, can you replace the line back up ?

mri/operators/fourier/non_cartesian.py
```diff
@@ -10,29 +10,30 @@
 """
 Fourier operators for cartesian and non-cartesian space.
 """
-
```
zaccharieramzi: let's avoid whiteline changes

chaithyagr:
Done, removing in progress tag

.travis.yml
```diff
@@ -40,26 +38,19 @@ install:
...
     - pip install coverage nose pytest pytest-cov
     - pip install coveralls
     - pip install pycodestyle
-    - pip install pybind11 nibabel pyqt5 pyqtgraph astropy 
-    - pip install git+https://github.com/CEA-COSMIC/ModOpt.git
-    - pushd ../
-    - git clone https://github.com/CEA-COSMIC/pysap
-    - cd pysap
-    - python setup.py install
+    - pip install -b pysap_build --no-clean --no-deps git+https://github.com/cea-cosmic/pysap
```
zaccharieramzi: why don't we want the dependencies of `pysap`?

mri/operators/fourier/non_cartesian.py
```diff
@@ -13,26 +13,28 @@
 
 # System import
 import warnings
-import numpy as np
```
zaccharieramzi: Since we changed the distribution is it really necessary to change the import order?

chaithyagr: Hmm, I can change it back, but it would help to keep as a safeguard in my opinion

.travis.yml
chaithyagr: Oh my bad, bad copy, changing this. I needed noclean

chaithyagr: And thats why the tests are failing

chaithyagr:
I just force pushed 2 changes. We still have issues with latest scipy version and this must be fixed. But that in my opinion is a bug in Modopt and I will open the same

PR #65, opened 2020-01-10
Move back tests to Right locations

chaithyagr:
This tackles moving back tests to right locations


chaithyagr:
@zaccharieramzi This is just moving tests, I think we can merge directly. Currently they fail till #63 goes through

PR #73, opened 2020-01-22
Fix circleci

chaithyagr:
This is a WIP PR to fix Circle CI build.
It looks like CI is running with very old pip. I have upgraded it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,7 @@ jobs:
           command: |
             python3 -m venv venv
             . venv/bin/activate
+            pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


chaithyagr:
I see that build is passing for me at https://circleci.com/gh/chaithyagr/pysap-mri/tree/fix_circleci

@zaccharieramzi this is a small change and can we merge this to see if this fixes the issue?
It looked like we were using old pip in the organization while a new one in my local CI builds. 
Now we run with pip 20, and things seem to be smooth.
Also I took the oppurtunity to add back the stack example which wasnt running earlier.

Issue #77, opened 2020-01-29
Gradient operator not initialized in SelfCalibrationReconstructor when sensivity maps are given manually


Daval-G:
I am using a `SelfCalibrationReconstructor` on the same data over and over and I would like to not compute the sensivity maps everytime but save them once and then give them to the reconstructor. 

The `reconstruct` method already has an argument to avoid recomputing the sensivity maps here:
https://github.com/CEA-COSMIC/pysap-mri/blob/c39764bf07c8dbefdb60c4e0be92fa3f96b477e2/mri/reconstructors/self_calibrating.py#L129

The problem is that currently the reconstructor does not initialize the gradient operator during its own initialization:
https://github.com/CEA-COSMIC/pysap-mri/blob/c39764bf07c8dbefdb60c4e0be92fa3f96b477e2/mri/reconstructors/self_calibrating.py#L111

So, the only way the `SelfCalibrationReconstructor` can initialize the gradient operator is by letting it recompute the sensivity maps at least once there:
https://github.com/CEA-COSMIC/pysap-mri/blob/c39764bf07c8dbefdb60c4e0be92fa3f96b477e2/mri/reconstructors/self_calibrating.py#L174

chaithyagr:
Marking this as an enhancement for now. @Daval-G are you doing this? Assigning you for now.
Also, I was wondering if we could get a verbosity that shows the Smaps after computation? Is this doable in this PR @Daval-G ?

Daval-G:
I will do a PR to fix that problem soon. Concerning the verbosity feature I don't think it is the job of the reconstructor to do so. But for sure there is a need for a better way to retrieve the Smaps than `smaps = reconstructor.extra_grad_args["Smaps"]`, so it would make sense to either add a documented getter, or simply to return the smaps along with `x_final,` `costs` and `metrics`. I prefer the first option.

chaithyagr:
The latter is surely problematic as that means specifically to this reconstructor we have different output. 
I agree with a getter function for obtaining `Smaps`. 
Maybe a setter function also that re-initilizes the gradient when `Smaps` is updated?

Daval-G:
It makes sense to me. I will include the getter and setter in my future PR.

chaithyagr:
Merged. in #79 , closing

PR #79, opened 2020-02-03
Extend SelfCalibrationReconstructor smaps use, add getter and setter

Daval-G:
This fixes issue #77 by adding sensivity maps as an optional argument to `SelfCalibrationReconstructor` during initialization. It also adds a getter and a setter for those sensivity maps with the appropriate documentation, in order to retrieve more easily the maps after a long computation.

The capital letter of the `Smaps` argument was kept to be coherent with the gradient operator `Smaps` argument, and to avoid unexpected behaviors by for example giving `smaps` as regular argument and `Smaps` in kwargs during `SelfCalibrationReconstructor` initialization.

### DIFF
--- a/mri/reconstructors/self_calibrating.py
+++ b/mri/reconstructors/self_calibrating.py
@@ -55,14 +55,21 @@ class SelfCalibrationReconstructor(ReconstructorBase):
         int or tuple indicating the k-space portion used to estimate the coil
         sensitivity information.
         if int, will be evaluated to (0.1,)*nb_dim of the image
+    Smaps: np.ndarray (optional, default None)
+        Sensivity maps used to initialize the gradient operator. If set to None,
+        the maps will have to be recomputed once when calling the reconstruct
+        method. The shape should correspond to the shape of the expected volume.
+        for gradient initialization:
+            Please refer to mri.operators.gradient.base for information
     smaps_extraction_mode: string 'FFT' | 'NFFT' | 'Stack' | 'gridding' default
         Defines the mode in which we would want to interpolate to extract the
-        sensitivity information,
+        sensitivity information when recomputing the sensivity maps.
         NOTE: FFT should be considered only if the input has
         been sampled on the grid
     smaps_gridding_method: string 'linear' (default) | 'cubic' | 'nearest'
         For gridding mode, it defines the way interpolation must be done used
-        by the sensitivity extraction method.
+        by the sensitivity extraction method when recomputing
+        the sensibity maps.
     n_jobs: int, default 1
         The number of CPUs used to accelerate the reconstruction
     verbose: int, optional default 0
@@ -83,7 +90,7 @@ class SelfCalibrationReconstructor(ReconstructorBase):
 
     def __init__(self, fourier_op, linear_op=None,
                  gradient_formulation="synthesis", kspace_portion=0.1,
-                 smaps_extraction_mode='gridding',
+                 Smaps=None, smaps_extraction_mode='gridding',
                  smaps_gridding_method='linear', n_jobs=1, verbose=0,
                  **kwargs):
         if linear_op is None:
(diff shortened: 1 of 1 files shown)


mri/reconstructors/self_calibrating.py
```diff
@@ -125,6 +135,30 @@ def __init__(self, fourier_op, linear_op=None,
...
 
+    def get_smaps(self):
+        """ This method returns the sensivity maps.
+
+        Returns
+        -------
+        np.ndarray, None
+            The sensivity maps when given or already computed, or None.
+        """
+
```
chaithyagr: Do we need this white space? Lets try to limit unwanted white spaces

mri/reconstructors/self_calibrating.py
```diff
@@ -170,8 +204,7 @@ def reconstruct(self, kspace_data, optimization_alg='pogm', x_init=None,
                 method=self.smaps_gridding_method,
                 n_cpu=self.n_jobs
             )
-            self.extra_grad_args['Smaps'] = Smaps
-            self.initialize_gradient_op(**self.extra_grad_args)
+            self.set_smaps(self, Smaps)
```
chaithyagr: If we are anyway using setters and getters why not do it the right way? We can that way make Smaps as a property of class with the functions as defined by you as setter and getter... 
You can refer https://medium.com/@pranaygore/setters-and-getters-in-python-76b5473b3c83 or anything similar..

chaithyagr:
Also @Daval-G , the Travis tests are failing for 2 reasons:
1) Bad usage of set_smaps()
2) Bad docstrings due to which docs couldnt be rightly generated.
Can you please look into it? (You can find logs here: https://travis-ci.org/CEA-COSMIC/pysap-mri/builds/645384188?utm_source=github_status&utm_medium=notification)

mri/reconstructors/self_calibrating.py
Daval-G: I wouldn't describe that as "the right way". Even the non-canonical source you refer to describes it as an optional way, and the PEP8 does not specify how to handle getters and setters. I am myself not a big fan of this syntax because the user has no way to know if s.he is accessing the attribute correctly.

But for sure we can store the `Smaps` as an explicit class attribute instead of a value stored in `self.extra_grad_args,` but currently there is no other use for it, and we would have to store it in `self.extra_grad_args` anyway. Also some may prefer `setSmaps` instead of `set_smaps`, since it also matches the capital letter of the variable.

chaithyagr: > but currently there is no other use for it, and we would have to store it in self.extra_grad_args anyway. 

Do we have to save it anyway even if we have the attribute like self._smaps? Why cant we explictly call as follows:

`self.initialize_gradient_op(**self.extra_grad_args, Smaps=self.smaps)` 

@zaccharieramzi do you have any thoughts on the best way to have a setter and getter?

The idea here is that we want to handle the following cases (please add if I missed anything):

1) Reinitialize gradient operator when Fourier operator changes (smaps are the same)
2) Get new smaps (maybe through other methods), and initialize gradient operator with them
3) Maybe just reconstruct for different images, with the same Smaps.

Daval-G: > Do we have to save it anyway even if we have the attribute like self._smaps? Why cant we explictly call as follows:
> 
> self.initialize_gradient_op(**self.extra_grad_args, Smaps=self.smaps)

It means we would have to systematically align the default value of `Smaps` in the `SelfCalibrationReconstructor` to the one in the gradient operator initializer. We avoid that by just not adding anything to the `self.extra_grad_args` variable when not needed. That won't necessarily be an issue in the future but that something to keep in mind.

More generally I don't see the point of letting the user change every piece of the reconstructor when it's computationally cheap to just instantiate another reconstructor. The idea of this PR was to:

1. Be able to give acquired/pre-computed Smaps to the reconstructor instead of letting the reconstructor compute them at least once.
2. Give to the user a proper way to retrieve the Smaps in the case they were computed by the reconstructor (to later re-run the script without having to compute them thanks to 1.)

zaccharieramzi: I don't necessarily see a point in having the getters or setters in one way or another in this particular case, so I would say let's go with what @Daval-G has implemented which does the job described [here](https://github.com/CEA-COSMIC/pysap-mri/pull/79#discussion_r375354520).

chaithyagr:
Looks like everything is resolved, @Daval-G please Squash and Merge when tests pass and everything is green

PR #82, opened 2020-02-10
Gridding Stack, add interpolation in z

### LINKED ISSUE #81: Fix Stack3DNFFT for VDS case to interpolate in z direction so that we do not zero fill
Currently, the Stack3DNFFT works as follows:
1) 2D NFFT to get planes at appropriate locations
2) 1D FFT along z

However, we need an extra step to interpolate samples in between locations especially for the case of VDS along z-direction. 
Ideally, this PR should work on Stacked3DNFFT : 

https://github.com/CEA-COSMIC/pysap-mri/blob/7b6b1236f4461b5e5ecedf54f6da912d0cdafba7/mri/operators/fourier/non_cartesian.py#L545-L552

However, I find same issue with gridding based reconstruction:

https://github.com/CEA-COSMIC/pysap-mri/blob/7b6b1236f4461b5e5ecedf54f6da912d0cdafba7/mri/operators/fourier/utils.py#L246-L253

This function is used in Stack based estimation of Smaps, and thus will severely impact the reconstruction results, hence both need to be fixed, however, I am not sure if Gridding in plan followed by gridding along z is the best case approximation to gridding in 3D at this point, will implement and see results

chaithyagr:
This resolves #81 

In this implementation, we have added checks to ensure that the kspace data is interpolated along z if we have undersampled in z, before actually carrying out IFFT.

This way, we are better approximating the kspace 3D gridding by 2D gridding and interpolation along z rather than simple zero filling.

### DIFF
--- a/mri/operators/fourier/utils.py
+++ b/mri/operators/fourier/utils.py
@@ -149,6 +149,8 @@ def get_stacks_fourier(kspace_loc, volume_shape):
     sampled_stack_z_loc = np.unique(kspace_loc[:, 2])
 
     try:
+        # We use np.isclose rather than '==' as the actual z_loc comes
+        # from scanner binary file that has been limited to floats
         idx_mask_z = np.asarray([
             np.where(np.isclose(z_loc, full_stack_z_loc))[0][0]
             for z_loc in sampled_stack_z_loc
--- a/mri/operators/fourier/utils.py
+++ b/mri/operators/fourier/utils.py
@@ -216,12 +216,12 @@ def gridded_inverse_fourier_transform_stack(kspace_data_sorted,
     """
     This function calculates the gridded Inverse fourier transform
     from Interpolated non-Cartesian data into a cartesian grid. However,
-    the IFFT is done similar to Stacked Fourier transform. 
-    We expect the kspace data to be limited to a grid on z, we calculate 
+    the IFFT is done similar to Stacked Fourier transform.
+    We expect the kspace data to be limited to a grid on z, we calculate
     the inverse fourier transform by-
     1) Grid data in each plane (for all points in a plane)
     2) Interpolate data along z, if we have undersampled data along z
-    3) Apply an IFFT on the 3D data that was gridded.    
+    3) Apply an IFFT on the 3D data that was gridded.
 
     Parameters
     ----------
@@ -257,7 +257,7 @@ def gridded_inverse_fourier_transform_stack(kspace_data_sorted,
             method=method,
             fill_value=0,
         )
-    # Check if we have undersampled in Z direction, in which case, 
+    # Check if we have undersampled in Z direction, in which case,
     # we need to interpolate values along z to get a good reconstruction.
     if len(idx_mask_z) < volume_shape[2]:
         # Interpolate along z direction
--- a/mri/operators/fourier/utils.py
+++ b/mri/operators/fourier/utils.py
@@ -223,7 +223,7 @@ def gridded_inverse_fourier_transform_stack(kspace_data_sorted,
     the inverse fourier transform by-
     1) Grid data in each plane (for all points in a plane)
     2) Interpolate data along z, if we have undersampled data along z
-    3) Apply an IFFT on the 3D data that was gridded.
+    3) Apply an IFFT on the 3D data that was gridded and interpolated in z.
 
     Parameters
     ----------
--- a/mri/operators/fourier/utils.py
+++ b/mri/operators/fourier/utils.py
@@ -150,7 +150,8 @@ def get_stacks_fourier(kspace_loc, volume_shape):
 
     try:
         idx_mask_z = np.asarray([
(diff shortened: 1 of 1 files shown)


mri/operators/fourier/utils.py
```diff
@@ -150,7 +150,8 @@ def get_stacks_fourier(kspace_loc, volume_shape):
 
     try:
         idx_mask_z = np.asarray([
-            np.where(x == full_stack_z_loc)[0][0] for x in sampled_stack_z_loc
+            np.where(np.isclose(x, full_stack_z_loc))[0][0]
+            for x in sampled_stack_z_loc
```
zaccharieramzi: we should change `x` to something more readable while we are at it. I was confused as to what it meant.

mri/operators/fourier/utils.py
```diff
@@ -150,7 +150,8 @@ def get_stacks_fourier(kspace_loc, volume_shape):
 
     try:
         idx_mask_z = np.asarray([
-            np.where(x == full_stack_z_loc)[0][0] for x in sampled_stack_z_loc
+            np.where(np.isclose(x, full_stack_z_loc))[0][0]
```
zaccharieramzi: why was this changed ?

mri/operators/fourier/utils.py
```diff
@@ -251,6 +252,28 @@ def gridded_inverse_fourier_transform_stack(kspace_data_sorted,
             method=method,
             fill_value=0,
         )
+    if len(idx_mask_z) < volume_shape[2]:
```
zaccharieramzi: can you explain a bit more what's happening here both in the PR description and in the docs of this function?

Also add some comments in the code about what's happening, for example: what does this test mean?

mri/operators/fourier/utils.py
chaithyagr: x == full_stack_z_loc does not work always, as seen by @mathrip .
Sometimes, we have issues just because they are approximately equal and not exactly equal.

zaccharieramzi: Why is not always equal ?
I mean how are they generated ?
I have a presupposition that is visibly wrong about how these stack locations are generated.

mri/operators/fourier/utils.py
chaithyagr: I generally explain things in issues and add PRs trying to point to what is happening. https://github.com/CEA-COSMIC/pysap-mri/issues/81 explains pretty much what is needed.
Do you want me to repeat it here?

However, I will add about what is done and what tests are added

mri/operators/fourier/utils.py
chaithyagr: So the way we generate is : 
**Expected:**  By linspace, we know we have Nz kspace z_lines, we expect them to be on a grid.
**Actual:** This is obtained from a function coming in from SPARKLING code that actually estimates where the kspace z_line is at based on what was run on the scanner, this is actually more accurate. (Note that we limit the discrete steps while running on scanner an there also could be some minor shifts). 

All we do here is to actually find the undersampled kspace locations based on what is sampled vs what should have been sampled.


So in reality, if we know that it is limited to a grid, with minor change in locations, we can still use this way of gridding rather than 3D

zaccharieramzi: ok makes sense

mri/operators/fourier/utils.py
zaccharieramzi: Yes I think redundancy in project track-keeping is good to a certain extent.
Generally the PR indicates what was actually done compared to the first intuition presented in the issue.

mri/operators/fourier/utils.py
```diff
@@ -216,6 +217,11 @@ def gridded_inverse_fourier_transform_stack(kspace_data_sorted,
     This function calculates the gridded Inverse fourier transform
     from Interpolated non-Cartesian data into a cartesian grid. However,
     the IFFT is done similar to Stacked Fourier transform.
+    We expect the kspace data to be limited to a grid on z, we calculate
+    the inverse fourier transform by-
+    1) Grid data in each plane (for all points in a plane)
+    2) Interpolate data along z, if we have undersampled data along z
+    3) Apply an IFFT on the 3D data that was gridded.
```
zaccharieramzi: gridded then interpolated.

mri/operators/fourier/utils.py
chaithyagr: Yup, I covered that in PR. Will keep this in mind for future PRs

mri/operators/fourier/utils.py
chaithyagr: Yes, that is what I have written right? `1)` Is gridding on each plane, `2)` is the actual interpolation in z.
Do you want me to explicitly mention this again in point 2?

zaccharieramzi: yes yes, I just think it's better reminded in point 3

PR #74, opened 2020-01-23
Add Calibrationless Example

chaithyagr:
This PR adds calibrationless examples with:
1) GroupLASSO Regularizer : Needs PR https://github.com/CEA-COSMIC/ModOpt/pull/78 
2) OWL Norm Regularizer : Needs PR https://github.com/CEA-COSMIC/ModOpt/pull/80

PR #83, opened 2020-02-10
[SPLIT][SMALL] Fix cost to work more robustly for primal_dual optimization

chaithyagr:
This PR splits up PR #74 . 

We need this merged after which I will rebase #74

### DIFF
--- a/mri/optimizers/utils/cost.py
+++ b/mri/optimizers/utils/cost.py
@@ -145,5 +145,8 @@ class GenericCost(costObj):
         if self.optimizer_type == 'forward_backward':
             cost = self.gradient_op.cost(x_new) + self.prox_op.cost(x_new)
         else:
+            # In primal dual algorithm, the value of args[0] is the data in
+            # Wavelet Space, while x_new is data in Image space.
+            # TODO, we need to generalize this
             cost = self.gradient_op.cost(x_new) + self.prox_op.cost(args[0])
         return cost
--- a/mri/optimizers/utils/cost.py
+++ b/mri/optimizers/utils/cost.py
@@ -22,7 +22,7 @@ class DualGapCost(costObj):
     """ Define the dual-gap cost function.
     """
     def __init__(self, linear_op, initial_cost=1e6, tolerance=1e-4,
-                 cost_interval=1, test_range=4, verbose=False,
+                 cost_interval=None, test_range=4, verbose=False,
                  plot_output=None):
         """ Initialize the 'DualGapCost' class.
         Parameters
@@ -36,7 +36,8 @@ class DualGapCost(costObj):
         tolerance: float, optional
             Tolerance threshold for convergence (default is "1e-4")
         cost_interval: int, optional
-            Iteration interval to calculate cost (default is "1")
+            Iteration interval to calculate cost (default is "None")
+            if None, cost is never calculated.
         test_range: int, optional
             Number of cost values to be used in test (default is "4")
         verbose: bool, optional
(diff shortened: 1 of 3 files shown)


chaithyagr:
The CircleCI is failing as Modopt release has not yet been done..

chaithyagr:
With new Modopt, we are now passing!!

mri/optimizers/utils/cost.py
```diff
@@ -133,5 +142,8 @@ def _calc_cost(self, x_new, *args, **kwargs):
         cost: float
             the cost function defined by the operators (gradient + prox_op).
         """
-        cost = self.gradient_op.cost(x_new) + self.prox_op.cost(x_new)
+        if self.optimizer_type == 'forward_backward':
+            cost = self.gradient_op.cost(x_new) + self.prox_op.cost(x_new)
+        else:
+            cost = self.gradient_op.cost(x_new) + self.prox_op.cost(args[0])
```
zaccharieramzi: are we sure that `args[0]` is always going to be the linear op applied to the optimisation variable ?

in any case it's not very clear hear and deserves at least a comment, but I would be happy if there was a better way.

mri/reconstructors/base.py
```diff
@@ -138,8 +137,16 @@ def reconstruct(self, kspace_data, optimization_alg='pogm',
...
+            optimizer_type = 'primal_dual'
         else:
             kwargs["prox_op"] = self.prox_op
+            optimizer_type = 'forward_backward'
+        self.cost_op = GenericCost(
+            gradient_op=self.gradient_op,
+            prox_op=self.prox_op,
+            verbose=self.verbose >= 20,
+            optimizer_type=optimizer_type
+        )
```
zaccharieramzi: you are not using `cost_op_kwargs`

mri/optimizers/utils/cost.py
```diff
@@ -22,7 +22,7 @@ class DualGapCost(costObj):
     """ Define the dual-gap cost function.
     """
     def __init__(self, linear_op, initial_cost=1e6, tolerance=1e-4,
-                 cost_interval=1, test_range=4, verbose=False,
+                 cost_interval=None, test_range=4, verbose=False,
```
zaccharieramzi: I feel like this is unrelated

chaithyagr: This is just to speed up reconstruction as currently, we calculate the cost at every iteration.

It is related to this PR, only in the sense that we are changing cost related stuff here. 
It is just a single line change.

zaccharieramzi: Makes sense.

mri/optimizers/utils/cost.py
chaithyagr: I agree that this is bad and non-scalable code, but for now I could not find a better way to do it. 
I have added a TODO to generalize the code here, also added a comment explaining the code better

zaccharieramzi: ok sure

mri/reconstructors/base.py
```diff
@@ -102,14 +102,10 @@ def initialize_gradient_op(self, **extra_args):
...
         )
-        self.cost_op = GenericCost(
-            gradient_op=self.gradient_op,
-            prox_op=self.prox_op,
-            verbose=self.verbose >= 20,
-        )
 
     def reconstruct(self, kspace_data, optimization_alg='pogm',
-                    x_init=None, num_iterations=100, **kwargs):
+                    x_init=None, num_iterations=100, cost_op_kwargs={},
```
zaccharieramzi: generally we should avoid mutable as defaults, so you could have `None` here and test `if cost_op_kwargs is None: cost_op_kwargs = {}`

PR #87, opened 2020-02-19
Add Support for gpuNUFFT

chaithyagr:
This PR adds support to gpuNUFFT. 

Currrent, GPU implementation of NUFFT is pyNUFFT is comparitively slower and also not very memory efficient, and also faces some memory leaks. 

The plan is to have this implementation right now and slowly move out from using pynufft, once this is stable.

@mathrip @Daval-G Heres how you can use this right away:

1) Install gpuNUFFT Library :
    * Clone https://github.com/chaithyagr/gpuNUFFT/ to branch `python_support`
    * Run `python setup.py install`
    * This must be pretty straight forward, let me know if you are facing issues in this.
 
2) Use gpuNUFFT ! 
    * Just call NonCartesianFFT with `implementation='gpuNUFFT'` 

I have done adjoint operation tests on 2D and 3D and also multichannel, however it is local and I have added code updates in `tests_local` for this.

@zaccharieramzi I am adding you for review of codes and documentation:
However, I am not adding gpuNUFFT installation in this PR, as the installation could be complicated based on presence of CUDA and MATLAB (my PR on gpuNUFFT builds both python bindings and MATLAB bindings, I still need to separate out the installations) 

Although I cant add @Daval-G and @mathrip to review, being users, I hope you can give valuable feedback about stability and cases that I missed in code.


chaithyagr:
Adding @Daval-G as official reviewer to this. 
@Daval-G please ensure the code is sane by running any examples. Feel free to suggest enhancements to both codes and also tests.
We will rely on @zaccharieramzi for final review before merge.

chaithyagr:
@zaccharieramzi I agree to that. 
In fact now, `cuda` as an earlier option does not make sense. This is because even gpuNUFFT is written in plain CUDA (while pynufft interfaces with help of reikna).
In my opinion, we can call earlier methods as `pynufft-cuda` and `pynufft-opencl` and new method as `gpuNUFFT`

However, I didnt want to make these changes as it will affect some legacy codes and this move must be more gradual, and to be honest, I feel the best thing to do is remove pyNUFFT slowly as the only advantage it offers is openCL version of code.

chaithyagr:
Added documentation and also updated API to new flow. I will wait for this to pass and merge this. In meantime @zaccharieramzi do let me know if there are any issues..

mri/operators/fourier/non_cartesian.py
```diff
@@ -34,6 +34,13 @@
                   "the master release. Till then you cannot use NUFFT on GPU")
     pass
 
+try:
+    from gpuNUFFT import NUFFTOp
+except ImportError:
+    warnings.warn("gpuNUFFT python package has not been found. If needed use "
+                  "the master release.")
```
zaccharieramzi: what do you mean by that?

mri/operators/fourier/non_cartesian.py
```diff
@@ -34,6 +34,13 @@
...
+    pass
```
zaccharieramzi: why this `pass`?

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +385,127 @@ def adj_op(self, x):
...
+            warnings.warn("Samples will be normalized between [-0.5; 0.5[")
+            samples = normalize_frequency_locations(samples)
+            self.samples = samples
+        if density_comp is None:
+            density_comp = np.ones(samples.shape[0])
+        if smaps is not None:
+            self.uses_smaps = False
+        else:
+            self.uses_smaps = True
+        self.operator = NUFFTOp(
```
zaccharieramzi: I guess this will raise a `NotFoundError`. Maybe it would be better when doing the import to have a flag set. This way we can give a better error message here.

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +385,127 @@ def adj_op(self, x):
...
 
+class gpuNUFFT:
+    """  GPU implementation of N-D non uniform Fast Fourrier Transform class.
+
+    Attributes
+    ----------
+    samples: np.ndarray
+        the mask samples in the Fourier domain.
+    shape: tuple of int
+        shape of the image (necessarly a square/cubic matrix).
```
zaccharieramzi: that's a shame, where does this requirement come from?

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +385,127 @@ def adj_op(self, x):
         return img * np.sqrt(np.prod(self.Kd))
 
 
+class gpuNUFFT:
+    """  GPU implementation of N-D non uniform Fast Fourrier Transform class.
+
+    Attributes
+    ----------
+    samples: np.ndarray
+        the mask samples in the Fourier domain.
```
zaccharieramzi: Are these the locations or the values? I am asking because there is a discrepancy with the `NFFT` class. I just saw that in `NUFFT` it's a bit different, but we still should address it and make that clear.

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +385,127 @@ def adj_op(self, x):
...
+            input array with the same shape as shape.
+
+        Returns
+        -------
+        np.ndarray
+            Non Uniform Fourier transform of the input image.
+        """
+        if self.n_coils > 1:
+            coeff = self.operator.op(np.asarray(
+                [np.reshape(image_ch.T, image_ch.size) for image_ch in image]
```
zaccharieramzi: I don't understand this transpose-reshape. Can you add a comment on it?

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +385,127 @@ def adj_op(self, x):
...
+            ).T)
+        else:
+            coeff = self.operator.op(np.reshape(image.T, image.size))
+            coeff = coeff[0]
+        return np.squeeze(coeff)
+
+    def adj_op(self, coeff):
+        """ This method calculates inverse masked non-uniform Fourier
```
zaccharieramzi: this is not the inverse right?

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +385,127 @@ def adj_op(self, x):
...
+            Non Uniform Fourier transform of the input image.
+        """
+        if self.n_coils > 1:
+            coeff = self.operator.op(np.asarray(
+                [np.reshape(image_ch.T, image_ch.size) for image_ch in image]
+            ).T)
+        else:
+            coeff = self.operator.op(np.reshape(image.T, image.size))
+            coeff = coeff[0]
+        return np.squeeze(coeff)
```
zaccharieramzi: why do you need to squeeze at the end of both ops?

mri/operators/fourier/non_cartesian.py
```diff
@@ -391,11 +516,14 @@ def __init__(self, samples, shape, implementation='cpu', n_coils=1):
...
             shape of the image (not necessarly a square matrix).
-        implementation: str 'cpu' | 'cuda' | 'opencl', default 'cpu'
+        implementation: str 'cpu' | 'cuda' | 'opencl' | 'gpuNUFFT',
+        default 'cpu'
             which implementation of NFFT to use.
         n_coils: int default 1
             Number of coils used to acquire the signal in case of multiarray
             receiver coils acquisition
+        kwargs: extra keyword args
+            these arguments are passed to NUFFT operator
```
zaccharieramzi: in this cas only `gpuNUFFT`. This should be clarified.

mri/operators/fourier/non_cartesian.py
chaithyagr: Will update the comment and point to README

mri/operators/fourier/non_cartesian.py
chaithyagr: To ensure that not having gpuNUFFT library still allows the use of pysap-mri . The user can use CPU NFFT still

mri/operators/fourier/non_cartesian.py
chaithyagr: Makes sense, will add that

mri/operators/fourier/non_cartesian.py
zaccharieramzi: well you don't need a `pass` for that do you? the `except` should be enough

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +385,127 @@ def adj_op(self, x):
...
+        if samples.min() < -0.5 or samples.max() >= 0.5:
+            warnings.warn("Samples will be normalized between [-0.5; 0.5[")
+            samples = normalize_frequency_locations(samples)
+            self.samples = samples
+        if density_comp is None:
+            density_comp = np.ones(samples.shape[0])
+        if smaps is not None:
+            self.uses_smaps = False
+        else:
+            self.uses_smaps = True
```
zaccharieramzi: also, I forgot to say that you should comment on what `uses_smaps` is used for, as it is not in this class

mri/operators/fourier/non_cartesian.py
chaithyagr: Agreed, will change that

mri/operators/fourier/non_cartesian.py
chaithyagr: Oh damn, missed `not` :face_with_head_bandage:

mri/operators/fourier/non_cartesian.py
chaithyagr: Will update the doc

mri/operators/fourier/non_cartesian.py
chaithyagr: Yes I was wondering on how to comment on this.
This uses C++ bindings that have a different memory ordering, and this is why we need to do this. DOes this make sense? I will add this now

mri/operators/fourier/non_cartesian.py
chaithyagr: Adjoint* yes

mri/operators/fourier/non_cartesian.py
chaithyagr: The C++ is generalized for 2D and 3D and thus always returns Num_channel x Nx x Ny x Nz. Thus for 2D case it is very important to squeeze the data up as Nz would be 1. (Same for multichannel)

mri/operators/fourier/non_cartesian.py
chaithyagr: It is not used now, but can be used for future, but yes it doesnt make sense here, I will remove it.

zaccharieramzi: no no you use it in the reconstructor!

mri/operators/fourier/non_cartesian.py
zaccharieramzi: ok makes sense. I guess you don't need the `coeff[0]` just above then.

mri/operators/fourier/non_cartesian.py
chaithyagr: Yes I removed that too. Idea is to use in in reconstructor if the fourier op has the smap, we would not recalculate it. But we ont need to do this.

mri/operators/fourier/non_cartesian.py
```diff
@@ -34,6 +34,14 @@
                   "the master release. Till then you cannot use NUFFT on GPU")
     pass
 
+try:
+    from gpuNUFFT import NUFFTOp
+    gpunufft_available = True
```
zaccharieramzi: a neater way to do it is to have that in the `else` of the `try/except`. You generally want a single statement in the `try` to make sure you control the error source

mri/operators/fourier/non_cartesian.py
zaccharieramzi: hmm but can you add explicitly the original shape and the end shape. To me it's not clear how we can avoid just reverting to the original image size. There certainly is a mechanism in the `reshape` for that, but it's not explicit.

mri/operators/fourier/non_cartesian.py
zaccharieramzi: you didn't use the flag at the beginning of the call to `__init__`

mri/operators/fourier/non_cartesian.py
zaccharieramzi: or you don't anymore, I don't understand. How do you cope with the situation where sensitivity maps are not provided?

zaccharieramzi: Ok (sorry for the above I hadn't read your comment).

mri/operators/fourier/non_cartesian.py
chaithyagr: There is else for try except?? :o

mri/operators/fourier/non_cartesian.py
chaithyagr: Well for multichannel, I know it must be coeff[0] as Num_channels is 1. 
I just want to try to get away from squeezing as much as possible so that code is more readable

zaccharieramzi: sure ok. I say let's comment on all this in the code.

mri/operators/fourier/non_cartesian.py
chaithyagr: Ideally we would need order='F'
However, I don't think this worked as python seemed to add a layer over the ordering. So ordering 'F' and reshape the way I did seemed to give different results.
This is something I would surely want to explore more, I think I will add it as a TODO

mri/operators/fourier/non_cartesian.py
zaccharieramzi: yes the extended `try/except` statement is:
```python
try:
  one-liner statement
except (Error1, Error2):
  throw custom warning
else:
  in case you want to do something only when try has succeeded
finally:
  general clean-up, for example if you opened a file or what-not
```

chaithyagr: Hmmm

chaithyagr: So I can skip finally? Cause we dont exactly have any cleanup right?

mri/operators/fourier/non_cartesian.py
chaithyagr: I just realized that we dont need squeeze for this case, as we used coeff[0]. Sorry that I missed that. I have documented code in other places where I used it and also here for coeff[0]

mri/operators/fourier/non_cartesian.py
zaccharieramzi: yes yes you don't need `finally` in this case!

mri/test_local/test_fourier_adjoint_gpu.py
```diff
@@ -25,7 +25,7 @@ def setUp(self):
         """
         self.N = 64
         self.num_channels = [1, 16]
-        self.platforms = ['cuda', 'opencl']
+        self.platforms = ['gpuNUFFT']#, 'cuda', 'opencl']
```
zaccharieramzi: I guess this is just temporary

mri/operators/fourier/non_cartesian.py
zaccharieramzi: ok weird. Does this work even with non-square data?

chaithyagr: Yup

mri/operators/fourier/non_cartesian.py
```diff
@@ -378,9 +387,134 @@ def adj_op(self, x):
         return img * np.sqrt(np.prod(self.Kd))
 
 
+class gpuNUFFT:
+    """  GPU implementation of N-D non uniform Fast Fourrier Transform class.
+
+    Attributes
+    ----------
+    samples: np.ndarray
+        the normalized kspace location values in the Fourier domain.
```
zaccharieramzi: Can you also update in the `__init__` and specify normalized between what and what?

chaithyagr: But __init__ will not come up in the documentation .

chaithyagr: Added it for now still ..

zaccharieramzi: I think it depends which doc handler you are using. I Jupyter notebooks you have both.

PR #89, opened 2020-02-25
Fix 3D Gridding

### LINKED ISSUE #88: [SMALL] Parallel Gridding fails as `griddata` fails when input arrays are non-parallel
The issue seen is exactly as described in https://stackoverflow.com/questions/41191911/why-does-scipy-interpolate-griddata-fail-for-readonly-arrays

It looks like in the current methodology, the Parallel Module from joblib sends in Read-Only Arrays to Griddata function, which causes errors.

`ValueError: buffer source array is read-only`

chaithyagr:
This resolves #88 

This issue was resolved by adding memmap='r+' to the codes to actually allow the memories to be writable.


chaithyagr:
@Daval-G , Any updates? I was hoping to get this in for a release.

chaithyagr:
Yes it currently prevents parallel gridding

PR #92, opened 2020-03-23
Add gpu sense recon support from GPUNUFFT

### LINKED ISSUE #91: Add support for using SENSE Recon with gpuNUFFT
Currently we have added support for gpuNUFFT as we wrote the python wrappers for the same.
However gpuNUFFT allows for getting SENSE recon which can happen on GPU if the Smaps are provided. 

This is something really helpful and I think its best to add this support in pysap-mri as it would speed up Self Calibrating Reconstruction

chaithyagr:
This resolves #91 .

In this PR, I have added support for SelfCalibrating Reconstructor to actually utilize the Sense Recon on gpuNUFFT, to speed up reconstructions.

I have added a new example in GPU_examples folder, which can be run locally to test this


chaithyagr:
Developer Review : @zaccharieramzi  

User Review : @Daval-G and @mathrip 
Please check the example scripts on use case.

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
```diff
@@ -0,0 +1,123 @@
+"""
+Non Cartesian Self Calibrating Reconstruction
+====----=====================================
+
+Author: Chaithya G R
+
+In this tutorial we will reconstruct an MRI image from non cartesian kspace
+measurements.
```
zaccharieramzi: Maybe specify this a bit more (like what is the specificity of this example compared to `examples/non_cartesian_reconstruction.py`).

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
```diff
@@ -0,0 +1,123 @@
...
+
+Import neuroimaging data
+------------------------
+
+We use the toy datasets available in pysap, more specifically a 2D parallel MRI
+brain slice on 32 channels and the acquisition non cartesian scheme.
```
zaccharieramzi: can you specify which scheme?

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
```diff
@@ -0,0 +1,123 @@
...
+# View Input
+# image.show()
+# mask.show()
+
+#############################################################################
+# Generate the kspace
+# -------------------
+#
+# From the 2D brain slice and the acquisition mask, we retrospectively
+# undersample the k-space using a cartesian acquisition mask
```
zaccharieramzi: non-cartesian right?

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
```diff
@@ -0,0 +1,123 @@
...
+# We then reconstruct the zero order solution as a baseline
+
+# Get the locations of the kspace samples and the associated observations
```
zaccharieramzi: that's not what you are doing here

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
```diff
@@ -0,0 +1,123 @@
...
+# image_rec0.show()
+base_ssim = ssim(image_rec0, image)
+print('The Base SSIM is : ' + str(base_ssim))
+
+# Obtain SMaps
+Smaps, SOS = get_Smaps(
+    k_space=kspace_obs,
+    img_shape=fourier_op.shape,
+    samples=kspace_loc,
+    thresh=(0.01, 0.01),
```
zaccharieramzi: I realize that thresh is not necessarily a very good variable name, so maybe a comment here would not be too much to explain what this does.

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
```diff
@@ -0,0 +1,123 @@
...
+    method='linear',
+    n_cpu=-1,
+)
+# Setup Fourier Operator with SENSE
+fourier_op_sense = NonCartesianFFT(
+    samples=kspace_loc,
+    shape=image.shape,
+    n_coils=cartesian_ref_image.shape[0],
+    smaps=Smaps,
+    implementation='gpuNUFFT',
```
zaccharieramzi: highlight how simple and seamless using GPU is here with a comment.

mri/operators/fourier/non_cartesian.py
```diff
@@ -445,6 +445,13 @@ def __init__(self, samples, shape, n_coils=1, density_comp=None,
             self.samples = samples
         if density_comp is None:
             density_comp = np.ones(samples.shape[0])
+        if smaps is None:
```
zaccharieramzi: wasn't this needed before because you didn't test or am I getting something wrong?

mri/operators/fourier/utils.py
```diff
@@ -286,3 +284,29 @@ def gridded_inverse_fourier_transform_stack(kspace_data_sorted,
...
+        supported
+
+    Returns
+    -------
+    bool
+        True if SENSE recon is being used
+    """
+    from .non_cartesian import NonCartesianFFT, gpuNUFFT
+    if isinstance(fourier_op, NonCartesianFFT):
+        if isinstance(fourier_op.implementation, gpuNUFFT):
```
zaccharieramzi: you can put them on the same line if you keep this order.

mri/reconstructors/self_calibrating.py
```diff
@@ -195,18 +204,19 @@ def reconstruct(self, kspace_data, optimization_alg='pogm', x_init=None,
...
-                samples=self.fourier_op.samples,
-                thresh=self.kspace_portion,
-                min_samples=self.fourier_op.samples.min(axis=0),
-                max_samples=self.fourier_op.samples.max(axis=0),
-                mode=self.smaps_extraction_mode,
-                method=self.smaps_gridding_method,
-                n_cpu=self.n_jobs
-            )
-            self.set_smaps(Smaps)
+            if not check_if_fourier_op_uses_sense(self.fourier_op):
```
zaccharieramzi: you could put that on the same line as `recompute_smaps` right ? this would also make it easier to review

mri/operators/fourier/non_cartesian.py
chaithyagr: We internally in C code just passed Nine by default earlier. It is only now that we have started supporting this

mri/operators/fourier/utils.py
chaithyagr: No dude, I tried that, it fails for Cartesian case for some reason.
I am assuming you mean to use `and` 
In C the operands are not calculated , here it looks like it is

mri/reconstructors/self_calibrating.py
chaithyagr: Yes, I think you are right here, will change

mri/operators/fourier/utils.py
zaccharieramzi: I am very surprised because I thought the second operand was not if the first was false. For example if I try the following one-liner: `bool(False and list().not_exist())`, I don't have an error, just `False`. 

Furthermore, if I do the following:
```python
import numpy as np

from mri.operators.fourier.cartesian import FFT
from mri.operators.fourier.non_cartesian import NonCartesianFFT, gpuNUFFT

f = FFT(samples=np.array([[0,0]]), shape=(1,1))
bool(isinstance(f, NonCartesianFFT) and isinstance(f.implementation, gpuNUFFT))
```

I just get `False` not an error. I guess there was another problem here.

mri/operators/fourier/non_cartesian.py
zaccharieramzi: ok makes sense

mri/operators/fourier/utils.py
chaithyagr: Hmmm, Lemme run this again

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
```diff
@@ -0,0 +1,127 @@
...
+# image_rec0.show()
+base_ssim = ssim(image_rec0, image)
+print('The Base SSIM is : ' + str(base_ssim))
+
+# Obtain SMaps
+Smaps, SOS = get_Smaps(
+    k_space=kspace_obs,
+    img_shape=fourier_op.shape,
+    samples=kspace_loc,
+    thresh=(0.01, 0.01),    # The cutoff threshold in each kspace direction
```
zaccharieramzi: Is it in percent of the size or in radians?

chaithyagr: Its a number of scale from 0 to 0.5 (As related to kspace_max ie)

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
chaithyagr: Oops, missed this.

mri/operators/fourier/utils.py
chaithyagr: Yes, you were right, earlier, I had some other issue, it is working normal as expected now.

examples/GPU_Examples/SelfCalibrationReconstruction_NonCartesian_gpuNUFFT_SENSE.py
chaithyagr: I have made a change on documenting this, lemme know if it makes more sense now.

zaccharieramzi: ok so in radians.

chaithyagr:
I will wait for @Daval-G or @mathrip to actually successfully run things before merging to mainline

zaccharieramzi:
Yes I think it's needed.

PR #94, opened 2020-04-03
changed self calibration gradients op to not use loop

zaccharieramzi:
small efficiency gain
I didn't write an issue for this since this is so minor


chaithyagr:
Thank you Zac. This seems to have missed in efficiency, as I think maybe the code was kept readable and loops tend to be the first approach to a problem.
I dont this there is much to review here. I will just run codes, and for sake of documentation, just write down the change in speed after this update in example codes. By then I hope CI will be complete and I will merge it.

PR #96, opened 2020-04-30
Add SWI Reconstruction of data from multichannel

mathrip:
Fix some errors encountered when processing SPARKLING data + add function to process SWI contrast images. 

References : 
- E. Mark Haacke et al., Susceptibility weighted imaging (SWI), Magnetic Resonance in Medicine 2004, 52:612–618


chaithyagr:
@mathrip I have converted the PR to a draft for now, to prevent emails to others. Once you think the PR is complete in your opinion and ready for review, for this I recomment you ensure that the docs are building fine, and also that PEP8 errors are good, please click on Ready for Review.

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+    ----------
+    imgs: np.ndarray
+        The images reconstructed channel by channel [Nch, Nx, Ny, Nz]
+    Returns
+    -------
+    I: np.ndarray
+        The combination of all the channels in a complex valued [Nx, Ny, Nz]
+    """
+    # Compute first the virtual coil
+    nch, nx, ny, nz = imgs.shape
```
chaithyagr, line 19: This could be a problem if we want to work with say only 2D images? Is there anyway we could make this more generic?

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+    weights = np.sum(np.abs(imgs), axis=0)
+    weights[weights==0] = 1e-16
+    phase_reference = np.asarray([np.angle(np.sum(
+        imgs[ch].flatten())) for ch in range(nch)])
+    reference = np.asarray([(imgs[ch] / weights) / np.exp(1j * phase_reference[ch])
+                           for ch in range(32)])
+    virtual_coil = np.sum(reference, axis=0)
+    difference_original_vs_virtual = np.asarray(
```
chaithyagr, line 27: There is a faster way to do this by direct multiplication rather than use loops, check PR by Zac here https://github.com/CEA-COSMIC/pysap-mri/pull/94/files

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+        [np.conjugate(imgs[ch]) * virtual_coil for ch in range(nch)])
+    hanning_1d = np.expand_dims(np.hanning(np.minimum(nx,ny)), 1)
+    hanning_2d = np.fft.fftshift(np.dot(hanning_1d, hanning_1d.T))
+    hanning_Nd = np.tile(np.expand_dims(np.tile(hanning_2d, (nch, 1, 1)), -1), (1, 1, 1, nz))
```
chaithyagr, line 31: Maybe try to generate the hanning filter using a loop here? That way we could be general with dimension.

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+#functions
```
chaithyagr, line 4: Unnecessary here

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+def virtual_coil_reconstruction(imgs):
+    """
+    Calculate the combination of all the coils thanks to the virtual coil
```
chaithyagr, line 7: Thanks to virtual coil? I am not sure if this is a typo.
`Calculate the combination of all coils using virtual coil reconstruction` could be used?

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+    methods as defined:
+    Parameters
```
chaithyagr, line 9: Documentation needs a line before `Parameters` and `Returns`

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+    virtual_coil = np.sum(reference, axis=0)
+    difference_original_vs_virtual = np.asarray(
+        [np.conjugate(imgs[ch]) * virtual_coil for ch in range(nch)])
+    hanning_1d = np.expand_dims(np.hanning(np.minimum(nx,ny)), 1)
+    hanning_2d = np.fft.fftshift(np.dot(hanning_1d, hanning_1d.T))
+    hanning_Nd = np.tile(np.expand_dims(np.tile(hanning_2d, (nch, 1, 1)), -1), (1, 1, 1, nz))
+    # Removing the background noise via low pass filtering
+    difference_original_vs_virtual = np.fft.ifft2(np.fft.fft2(
+        difference_original_vs_virtual, axes=(1, 2)) * hanning_Nd, axes=(1, 2))
+    I = np.asarray([imgs[ch] * np.exp(1j * np.angle(
```
chaithyagr, line 35: A better name for this variable than `I`? 
I am assuming it is per channel image obtained based on phase?

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    nx, ny, nz = volume_shape
+
+    # create hanning filter 3D
+    hanning_1d = np.zeros([np.minimum(nx,ny),1])
+    hanning_1d[int(np.round(3 * nx / 10)):int(np.round(7 * nx / 10))] = \
+        np.expand_dims(np.hanning(int(np.round(7 * nx / 10))-int(np.round(3 * nx / 10))), 1)
+    hanning_2d = np.fft.fftshift(np.dot(hanning_1d, hanning_1d.T))
+    hanning_Nd = np.tile(np.expand_dims(hanning_2d, -1), (1,1,nz))
```
chaithyagr, line 58: Same as above, better use a loop to generate the hanning filter. Also, if you are anyway having this function, why not use it above?

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+        ax[0, 1].imshow(np.abs(pha_mask_raw_final[:, :, show_slice].T), cmap='gray', aspect='equal')
+        ax[0, 1].title.set_text('Phase Mask')
+        ax[0, 2].axis('off')
+        ax[0, 2].imshow(np.abs(swi_from_raw[:, :, show_slice].T), cmap='gray', aspect='equal')
+        ax[0, 2].title.set_text('SW Image')
+        ax[1, 2].axis('off')
+        ax[1, 2].title.set_text('SW Image (mIP)')
+        plt.show()
+
+    return swi_from_raw
```
chaithyagr, line 121: Are we sure we dont want any other images? Like phase / phase mask etc?

mri/utils/process_SWI
```diff
@@ -0,0 +1,127 @@
...
+        ----------
+        image_ch: np.ndarray
+            Images reconstructed per channels in shape [Nch, Nx, Ny, Nz] in 3D or [Nch, Nx, Ny] in 2D
+        Returns
+        -------
+        SWI: np.ndarray
+            SWI image
+    """
+    if len(imRef.shape) == 3 :
+        image=imRef[:,:,:,np.newaxis]
```
chaithyagr, line 74: I'm not very comfortable with this, let us see if we can get hanning filter generateable for an arbitrary dimension using loop? 
We could resort to this as a last resort.

mri/utils/process_SWI
mathrip: I will release a more general version. 
But to answer your question, if you look closer this is not a 3D hanning filter. There is no filtering in z direction, it is just 2D plan repeated.
Plus this filter do not cover the same portion of the kspace in both the functions, this is why originaly I kept them seperated

mri/utils/process_SWI
mathrip: Well, we want to be able to display the process as I did in the function. Should it be appart from the function, on a separate script?

mri/utils/process_SWI
mathrip: It is not working like this for all of these operation though. 
I don't think about a better way to do these operations without looping by channels.

mri/operators/fourier/utils.py
```diff
@@ -97,11 +97,11 @@ def normalize_frequency_locations(samples, Kmax=None):
     """
     samples_locations = np.copy(samples.astype('float'))
     if Kmax is None:
-        Kmax = 2*np.abs(samples_locations).max(axis=0)
+        Kmax = np.abs(samples_locations).max(axis=0)
     elif isinstance(Kmax, (float, int)):
         Kmax = [Kmax] * samples_locations.shape[-1]
     Kmax = np.array(Kmax)
-    samples_locations /= Kmax
+    samples_locations /= 2*Kmax
```
chaithyagr, line 104: Why do we need this change? It would be better to place Kmax to actually represent Kmax, ie 0.5 and not 1.
Lets not make changes if its unnecessary..

mri/operators/fourier/utils.py
```diff
@@ -143,8 +143,8 @@ def get_stacks_fourier(kspace_loc, volume_shape):
     # Find the mask used to sample stacks in z direction
     full_stack_z_loc = convert_mask_to_locations(
         np.ones(volume_shape[2]),
-    )[:, 0]
-    sampled_stack_z_loc = np.unique(kspace_loc[:, 2])
+    )[:, 0].astype('float16')
+    sampled_stack_z_loc = np.unique(kspace_loc[:, 2]).astype('float16')
```
chaithyagr, line 147: I am a bit apprehensive about why we moved to float16 for comparision. If the earlier code was not working, it would be better to change the threshold for comparision in `np.isclose` below.

mri/operators/fourier/utils.py
```diff
@@ -143,8 +143,8 @@ def get_stacks_fourier(kspace_loc, volume_shape):
...
 
     try:
         # We use np.isclose rather than '==' as the actual z_loc comes
```
chaithyagr, line 150: As the change to np.isclose is already done, maybe we can remove this comment @mathrip ? As it is a small change, maybe we can do it in this PR itself.

mri/reconstructors/self_calibrating.py
```diff
@@ -205,6 +205,7 @@ def reconstruct(self, kspace_data, optimization_alg='pogm', x_init=None,
         if recompute_smaps and \
                 not check_if_fourier_op_uses_sense(self.fourier_op):
             # Extract Sensitivity maps and initialize gradient
+            print('start Smaps extraction')
```
chaithyagr, line 208: Let us not have print statements without any verbosity, either remove this or add it based on Verbosity.

mri/reconstructors/self_calibrating.py
```diff
@@ -217,7 +218,9 @@ def reconstruct(self, kspace_data, optimization_alg='pogm', x_init=None,
                 n_cpu=self.n_jobs
             )
             self.set_smaps(Smaps)
+            print('smaps extracted')
```
chaithyagr, line 221: Same as above here

mri/reconstructors/self_calibrating.py
```diff
@@ -217,7 +218,9 @@ def reconstruct(self, kspace_data, optimization_alg='pogm', x_init=None,
...
         # Start Reconstruction
+        print('start reconstruction')
```
chaithyagr, line 223: And here. I know that this could sometimes be slow step, but any prints that we have needs to be shielded.

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+# http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.html
+# for details.
+##########################################################################
+
+import numpy as np
+
+
+def virtual_coil_reconstruction(imgs):
+    """
+    Calculate the combination of all coils using virtual coil reconstruction
```
chaithyagr, line 15: Do we have some refrence that we can point to here?

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+        in shape [Nch, Nx, Ny, Nz] for 3D or [Nch, Nx, Ny] for 2D
+
+    Returns
+    -------
+    img_comb: np.ndarray
+        The combination of all the channels in a complex valued
+        in shape [Nx, Ny, Nz] for 3D or [Nx, Ny] for 2D
+    """
+    img_sh = imgs.shape
+    # Compute first the virtual coi
```
chaithyagr, line 30: Typo `coil`

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+    weights = np.sum(np.abs(imgs), axis=0)
+    weights[weights == 0] = 1e-16
+    phase_reference = np.asarray([np.angle(np.sum(
+        imgs[ch].flatten())) for ch in range(img_sh[0])])
+    reference = np.asarray([(imgs[ch] / weights) / np.exp(
+        1j * phase_reference[ch]) for ch in range(32)])
```
chaithyagr, line 36: `for ch in range(32)` ?
Seems to be hardcoded here, we may need to pickup number of channels from input

chaithyagr: Also, I think we can still prevent loop here, lemme update here after checking something myself.

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+    virtual_coil = np.sum(reference, axis=0)
+    difference_original_vs_virtual = np.conjugate(imgs) * virtual_coil
+    # Hanning filtering in readout and phase direction
+    hanning = get_hanning_filter([img_sh[1], img_sh[2]])
+    if len(img_sh) > 3:
+        hanning = np.tile(np.expand_dims(np.tile(
```
chaithyagr, line 42: Cant we add this inside get_hanning_filter? 
Maybe we just send image shape and get a complete hanning filter?

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+            hanning, (img_sh[0], 1, 1)), -1), (1, 1, 1, img_sh[3]))
+    # Removing the background noise via low pass filtering
+    difference_original_vs_virtual = np.fft.ifft2(np.fft.fft2(
```
chaithyagr, line 45: Is it FFT2 even if it is 3D volume? I am not sure about this, We may need to check this, from what I understand, the noise removal is done using a hanning filter in the same dimensions as that of image.

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+
+    Parameters
+    ----------
+    volume_shape: np.ndarray
+        size of the resulting hanning filter.
+
+    Returns
+    -------
+    hanning: np.ndarray
+        the hanning filter
```
chaithyagr, line 64: `the frequency response of hanning filter`

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+    # 1. get unique complex image
+    imTot = virtual_coil_reconstruction(imRef)
+
+    # 2. apply hanning filter on readout&phase direction from a kspace portion
+    hanning_filt = np.zeros([imTot.shape[0], imTot.shape[1]])
+    p_x = [int(np.round(3 * imTot.shape[0] / 10)),
+           int(np.round(7 * imTot.shape[0] / 10))]
+    p_y = [int(np.round(3 * imTot.shape[1] / 10)),
+           int(np.round(7 * imTot.shape[1] / 10))]
+    hanning_filt[p_x[0]:p_x[1], p_y[0]:p_y[1]] = get_hanning_filter(
```
chaithyagr, line 102: Is this 30 percent and 70 percent from literature? If they can be changed, I would prefer to have it as percentage of center of kspace (here it is 40) and keep it variable as a parameter

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+    imTot_filt_phase = np.angle(imTot_homodyne)
+    pha_mask_raw = np.ones(imTot_filt_phase.shape)
+    pha_mask_raw[imTot_filt_phase > 0] = (
+            1 - imTot_filt_phase[imTot_filt_phase > 0] / np.pi)
+    pha_mask_raw_final = pha_mask_raw ** 5
+
+    # 4. compute SWI
+    swi = np.abs(imTot) * pha_mask_raw_final
+
+    return swi
```
chaithyagr, line 123: I would prefer if we have the mask also returned for any debug if needed.

mri/operators/fourier/utils.py
mathrip: I think we've done this change to fix the problem we had with rotating stacks and the fact it was changing the FOV.

mri/operators/fourier/utils.py
mathrip: Agree with that, I'll have a look to change the threshold

mri/operators/fourier/utils.py
chaithyagr: How would this solve the issue? The `sample_locations` would still be the same in both the cases, we merely moved when we divide by 2.

mri/utils/process_SWI
mathrip: This virtual coil function comes from Loubna's previous work (https://github.com/LElgueddari/pysap/blob/calibrationless_p_mri_reconstruction/pysap/plugins/mri/parallel_mri/utils.py)
There is no hanning filtering in the 3D dimension.

mri/utils/process_SWI
mathrip: They are fixed and won't change. This is how SWI is defined from the constructors

mri/utils/process_SWI
mathrip: Sure, I'll return the mask

mri/utils/process_SWI
mathrip: There is no hanning filtering in the partition direction (z), plus the hanning filter is only applied on a portion of the 2D kspace before to be extended in the third dimension.
If we want to keep the get_hanning_filter general, this part can't be include in the function because it is really specific to the SWI process.

mri/reconstructors/self_calibrating.py
mathrip: Ok I'll remove it, it was for my personal understanding

mri/operators/fourier/utils.py
mathrip: Yes, I agree

mri/operators/fourier/utils.py
mathrip: Because at the begining we `*2` was only applied in the case Kmax is none and we needed to apply it also when we give a specific kmax : 
`elif isinstance(Kmax, (float, int)):	    
        Kmax = 2 * [Kmax] * samples_locations.shape[-1]` 
And instead of multiplying by 2 in both case it was easier to just divide by 2 the final kmax.

chaithyagr: Hmm interesting. 
Makes sense.

chaithyagr:
@mathrip any updates? There are still some changes left based on code. Also, can you please rename your file to have a `.py` extension, and add the appropriate LICENSE related header information?

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+    hanning = np.hanning(vol_shape[0])
+    for i in range(1, dim):
+        hanning = np.expand_dims(hanning, -1)
+        hanning = np.dot(hanning, np.expand_dims(np.hanning(
+            vol_shape[i]), -1).T)
+    return hanning
+
+
+def get_SWI(imRef):
+    """ Process SWI images from complex reconstructed image per channels
```
chaithyagr, line 78: Can we please add the reference here to Haccke et. al (with link preferably)

chaithyagr: https://pubmed.ncbi.nlm.nih.gov/15334582/

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+                imTot) * np.fft.fftshift(hanning_filt))
+    # compute homodyne image
+    imTot_homodyne = imTot * np.exp(-1j * np.angle(imTot_filtered))
+
+    # 3. get phase mask
+    imTot_filt_phase = np.angle(imTot_homodyne)
+    pha_mask_raw = np.ones(imTot_filt_phase.shape)
+    pha_mask_raw[imTot_filt_phase > 0] = (
+            1 - imTot_filt_phase[imTot_filt_phase > 0] / np.pi)
+    pha_mask_raw_final = pha_mask_raw ** 5
```
chaithyagr, line 118: Please check the implementation in Haccke, we are missing negative phase mask.

Also, I would prefer to have this number `5` to be changeable and a parameter.

chaithyagr:
This is increasingly getting stale, @Daval-G can you please ensure you have all codes from here in your codes? In which case, we can close this perhaps?

chaithyagr:
> This is increasingly getting stale, @Daval-G can you please ensure you have all codes from here in your codes? In which case, we can close this perhaps?

@Daval-G ? Can we close this as you will have this in #115 ?

Daval-G:
> > This is increasingly getting stale, @Daval-G can you please ensure you have all codes from here in your codes? In which case, we can close this perhaps?
> 
> @Daval-G ? Can we close this as you will have this in #115 ?

I don't actually intend to add SWI-related code in #115. I feel that current PR should still be good to merge.

mri/utils/process_SWI
```diff
@@ -0,0 +1,123 @@
...
+# Distributed under the terms of the CeCILL-B license, as published by
+# the CEA-CNRS-INRIA. Refer to the LICENSE file or to
+# http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.html
+# for details.
+##########################################################################
+
+import numpy as np
+
+
+def virtual_coil_reconstruction(imgs):
```
Daval-G, line 13: The term "reconstruction" is disturbing in the pysap-mri context, when in fact this is a coil "combination" method.

chaithyagr: > I don't actually intend to add SWI-related code in #115. I feel that current PR should still be good to merge.

Ahh, than we really need someone to work on these codes, maybe since you use them, can you take over from here? I would just want to ensure that we have this at main branch. We can make changes to this branch or maybe get a new branch for this.

chaithyagr:
We can maybe close this PR as it is stale. @Daval-G can you handle adding up to date version of these features and an example so that we can use it (or does it already exist?) Closing this here..

PR #99, opened 2020-05-26
[SHORT] Pass linear_op to POGM

### LINKED ISSUE #98: [Enhancement][BUG]  Send Linear operator to POGM so that metrics can be called
Current implementation of POGM in pysap-mri does not pass the linear_op to the optimizer due to which it is internally set to `Identity`. 
However, as POGM carries out complete optimization on the wavelet coefficients, the SSIM metric calculations needs the adjoint operation which is performed here:

https://github.com/CEA-COSMIC/ModOpt/blob/f9aa7b6154bfe34bc79d94d3c7d03350d47eef4a/modopt/opt/algorithms.py#L1347-L1356

However, for this, the linear operator must be set right which isnt done right in current scenario.
This prevents faster iterations of retrospective studies and also a better grid search.

Tagging as a bug to just raise the importance of the issue.

chaithyagr:
This is a quick single line PR to ensure that metrics can be used right with POGM.
This resolves #98 .

Adding @zaccharieramzi for a quick review as he wrote POGM initially.


chaithyagr:
Looks like a random crash. I will squash merge once green. Thank you @zaccharieramzi for the review.

chaithyagr:
The failure is for Python 3.5, looks like Scikit-Learn is no longer supporting python 3.5 
Even the master branch is failing with same issue.

@zaccharieramzi shall we also remove travis from testing in python 3.5 and start testing on Python 3.8?
Maybe I could add this to current PR, as this is currently just a single line change.

zaccharieramzi:
@chaithyagr I say let's have another PR for this just to be clean

chaithyagr:
Sure, although this must not be happening, I see this as a separate issue in pysap. It looks like doing python setup.py install, is triggering easy_install, which isnt handling the dependencies well, mostly that too in Travis. Let me dig a bit deeper into this.

chaithyagr:
Merging again, as 2 CIs have passed

PR #103, opened 2020-06-10
Update OWL to avoid weights overflow with large volumes.

Daval-G:
I encountered a problem with the OWL proximity operator as the size carried by the argument "band_shape" was too big in the case of my volume (32 channels, 384 Nx, 384 Ny, 208 Nz).
It was causing an overflow in the modified function, and raising later a ValueError because the weights were not ordered. Just changing dtype=float32 to dtype=float64 fixes the issue.


chaithyagr:
2 CI pipelines passed. I will merge this now, I dont see the point in waiting for the CI to be fixed now, and make this PR stale.

PR #105, opened 2020-07-03
Fix Parallel wavelet run by having a queue of transformers

### LINKED ISSUE #104: [BUG] [URGENT] Parallel WaveletN has issues when n_jobs!=1
For Calibratinless reconstruction, we would have to carry out wavelet transform across all the channels. 
In order to speed things up, I set up the codes for running linear_op in parallel. 
However, this caused the following issue:

In the definition of WaveletN, we define just one linear operator:

https://github.com/CEA-COSMIC/pysap-mri/blob/db04fe2956adaf0b1e84bc58dc99e5f63664286e/mri/operators/linear/wavelet.py#L61-L62

Then we use this in parallel to obtain our images in `_op()` function:

https://github.com/CEA-COSMIC/pysap-mri/blob/db04fe2956adaf0b1e84bc58dc99e5f63664286e/mri/operators/linear/wavelet.py#L71-L77

In doing so, we are eventually using the same wavelet operator for all the channels in parallel which causes severe races. 
Due to this the overall image is bad and has artifacts like phase wrapping in magnitude image.

I will get a PR fixing this, but for now, users are requested to not use `n_jobs`
Also, this seems like a hole in testing, hence adding a testing tag and I will add a test for the same.

Thanks to @Daval-G in helping in identifying this major issue...

chaithyagr:
This PR resolves #104 

1) I have basically implemented a wavelet queue which is used when `n_jobs != 1` 
2) It looks like we have `get_coeff` and `set_coeff` functions in WaveletN, but we never used them, this seems to be stale, removing them to keep code simple
3) Also, I have updated the tests to ensure we dont miss such bugs in future. I have refactored some tests to ensure there is no redundancy when it comes to testing multichannel or single channel

Adding @Daval-G for testing the codes with calibration less reconstruction
Adding @zaccharieramzi for final sign off on code structure and any issues with documentation (although there are no major changes)

### DIFF
--- a/mri/operators/linear/wavelet.py
+++ b/mri/operators/linear/wavelet.py
@@ -20,6 +20,7 @@ from pysap.base.utils import flatten
 from pysap.base.utils import unflatten
 
 # Third party import
+import joblib
 from joblib import Parallel, delayed
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


chaithyagr:
Thanks @Daval-G , @zaccharieramzi at your convenience please have a look at the code style, etc, I will merge it soon for a release.

mri/operators/linear/wavelet.py
```diff
@@ -52,28 +53,39 @@ def __init__(self, wavelet_name, nb_scale=4, verbose=0, dim=2,
...
+                verbose=verbose,
+                dim=dim,
+                **kwargs)
+            )
         self.coeffs_shape = None
 
-    def get_coeff(self):
-        return self.transform.analysis_data
-
-    def set_coeff(self, coeffs):
```
zaccharieramzi, line 68: Where were these used?

chaithyagr: This was used earlier by user, but now we dont tend to use these. The idea back then in my opinion was to just set and get coefficients, which is now clearly done by op and adj_op anyway.

zaccharieramzi: ok cool

PR #107, opened 2020-07-06
Fix travis based on new pysap flow and add support to python3.8

### LINKED ISSUE #102: silght difference in documentation and code
https://github.com/CEA-COSMIC/pysap-mri/blob/1f94b5ab6f7234594de2e1482768cecb243021bd/examples/gridsearch_launch.py#L11

The grid search is done over mus in the range between 1e-8 and 1e-6 while the doc sais it's 1e-8 and 1e-9.

chaithyagr:
1) With recent PR, we no longer need to add the binary file location (/extern/bin etc) to $PATH. This saves a lot of time and thus, pysap can be installed directly with pip install.

2) Also, added dependency to an internal released pynfft2 as the original pynfft was not released for quite some time and we would like to proceed with a release and not wait on them

3) Add python 3.8 for testing and overall move to testing on xenial as that corresponds to a higher ubuntu. Move `time.clock()` calls to `time.perf_counter()` calls to support python3.8

4) Remove support for n_jobs in ISAP wavelets due to erratic behavior.

5) Add a minute single documentation fix for #102. This resolves #102 

Will open the PR for review once I have passing travis

### DIFF
--- a/mri/operators/linear/wavelet.py
+++ b/mri/operators/linear/wavelet.py
@@ -67,7 +67,7 @@ class WaveletN(OperatorBase):
         if n_proc < 0:
             n_proc = joblib.cpu_count() + self.n_jobs + 1
         # Create transform queue for parallel execution
-        for i in range(min(n_proc, self.n_coils)):
+        for i in range(min(n_proc, self.n_coils)+1):
             self.transform_queue.append(transform_klass(
                 nb_scale=self.nb_scale,
                 verbose=verbose,
--- a/mri/operators/linear/wavelet.py
+++ b/mri/operators/linear/wavelet.py
@@ -18,11 +18,13 @@ from modopt.signal.wavelet import get_mr_filters, filter_convolve
 import pysap
 from pysap.base.utils import flatten
 from pysap.base.utils import unflatten
+from pysap.utils import wavelist
 
 # Third party import
 import joblib
 from joblib import Parallel, delayed
 import numpy as np
+import warnings
 
 
 class WaveletN(OperatorBase):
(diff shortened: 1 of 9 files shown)


mri/operators/linear/wavelet.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, wavelet_name, nb_scale=4, verbose=0, dim=2,
         if n_proc < 0:
             n_proc = joblib.cpu_count() + self.n_jobs + 1
         # Create transform queue for parallel execution
-        for i in range(min(n_proc, self.n_coils)):
+        for i in range(min(n_proc, self.n_coils)+1):
```
zaccharieramzi, line 70: what is this?

mri/optimizers/forward_backward.py
```diff
@@ -63,7 +63,7 @@ def fista(gradient_op, linear_op, prox_op, cost_op,
     metrics: dict
         the requested metrics values during the optimization.
     """
-    start = time.clock()
+    start = time.perf_counter()
```
zaccharieramzi, line 66: I am not necessarily against those test changes but maybe just reflect those in the PR title and description

mri/operators/linear/wavelet.py
chaithyagr: No , I seem to have minute issues with travis , I was getting some `None` for some wavelets. This was just a way to test that, I am debugging travis. Removing this.

mri/optimizers/forward_backward.py
chaithyagr: Will do it, this change had to be done as python3.8 has deprecated `time.clock()`

mri/operators/linear/wavelet.py
```diff
@@ -66,6 +68,11 @@ def __init__(self, wavelet_name, nb_scale=4, verbose=0, dim=2,
         n_proc = self.n_jobs
         if n_proc < 0:
             n_proc = joblib.cpu_count() + self.n_jobs + 1
+        if n_proc > 0:
+            if wavelet_name in wavelist()['isap-2d'] or wavelet_name in wavelist()['isap-3d']:
+                warnings.warn("n_jobs is currently unsupported for ISAP wavelets, setting n_jobs=1")
+                self.n_jobs = 1
+                n_proc = 1
         # Create transform queue for parallel execution
```
chaithyagr, line 76: I have removed support for n_jobs in ISAP wavelets as they are erratic in behaviour and crash randomly sometimes. 
We can get back at this once it is fixed in https://github.com/CEA-COSMIC/pysap/issues/101

zaccharieramzi: hmmm it's weird.
Can you document this (i.e. write the traceback and minimal reproducing example) in the issue?

chaithyagr: I have added the both in the pysap issue earlier

chaithyagr:
Yes, the circle CI is passing as it is testing on a cache, surprisingly, there is no way to force delete a cache, I have requested the support for this. Will merge soon..

PR #111, opened 2020-12-16
[SHORT] Fix up codes for FFT class, making it more user friendly and fix grid search

chaithyagr:
This PR Covers:
1. Option for user to send mask directly
2. Using `axes` parameter of `np.fft.fftn` to do FFT in vectorized way, than using loops
3. Send `n_coils` to parameters in grid search


mri/operators/fourier/cartesian.py
```diff
@@ -88,18 +95,26 @@ def op(self, img):
...
             if self.n_coils > 1 and self.n_coils != img.shape[0]:
                 raise ValueError("The number of coils parameter is not equal"
                                  "to the actual number of coils, the data must"
                                  "be reshaped as [n_coils, Nx, Ny, Nz]")
             else:
-                # TODO: Use joblib for parallelization
-                return np.asarray([self._mask * np.fft.ifftshift(np.fft.fftn(
-                                    np.fft.fftshift(img[ch]), norm="ortho"))
-                                   for ch in range(self.n_coils)])
+                axes = tuple(np.arange(1, img.ndim))
```
zaccharieramzi, line 106: I find this weird.
If for example `img` has 2 dimension, this will be `np.arange(1, 2)`, giving `array([1])`, which is not what we want, either for the `fftn` or for the `fftshift`.

Or am I missing something?

mri/operators/fourier/cartesian.py
```diff
@@ -88,18 +95,26 @@ def op(self, img):
...
+                return self.mask * np.fft.ifftshift(
+                    np.fft.fftn(
```
zaccharieramzi, line 108: Did you check (and if so how) that `fftn` is indeed faster than the loop?

Indeed I had this problem, summarized in this [SO question](https://stackoverflow.com/questions/58009758/numpy-fftn-very-inefficient-for-2d-fft-of-several-images) a long time ago.

Finally, it might be nice to consider some parallelization on this (maybe for now as a TODO).

zaccharieramzi (changes requested):
Some questions

mri/operators/fourier/cartesian.py
chaithyagr: Well, there was a TODO for parallelization, but then I assumed that the fftn through numpy might as well handle this well and  in practice this should not be that big a gain, but you do have a valid point. I will check this and get back here.

mri/operators/fourier/cartesian.py
chaithyagr: ndim = dimension of image + 1 for multichannel case.

zaccharieramzi: ofc makes sense

mri/operators/fourier/cartesian.py
chaithyagr: moved to scipy

chaithyagr: Also added n_jobs and connected the argument to workers. This way, user has control on exact speed

mri/operators/fourier/cartesian.py
```diff
@@ -45,32 +46,42 @@ class FFT(OperatorBase):
...
-                 data shape must be equal to [n_coils, Nx, Ny, NZ]
+        n_coils: int, default 1
+            Number of coils used to acquire the signal in case of
+            multiarray receiver coils acquisition. If n_coils > 1,
+            data shape must be equal to [n_coils, Nx, Ny, NZ]
+        samples: np.ndarray, default None
+            the mask samples in the Fourier domain.
+        mask: np.ndarray, default None
+            the mask as a matrix with 1 at sample locations
+            please pass samples or mask
```
zaccharieramzi, line 71: Maybe you can also add a doc on `n_jobs`, especially specifying that it can be `-1` for all CPUs.

mri/operators/fourier/cartesian.py
```diff
@@ -45,32 +46,42 @@ class FFT(OperatorBase):
...
         """
-        self.samples = samples
```
zaccharieramzi, line 67: One test is failing because this is missing, can you check it out?

zaccharieramzi (changes requested):
One other test is failing for a matching problem, you should investigate to see if it's serious or not

mri/operators/fourier/cartesian.py
chaithyagr: I thought I did add info on that. Let me check

mri/operators/fourier/cartesian.py
chaithyagr: Yes, need to rebase i guess

mri/operators/fourier/cartesian.py
zaccharieramzi: yes it's here now

mri/operators/fourier/cartesian.py
zaccharieramzi: cool, let's see how the tests do

chaithyagr:
Looks like scipy for some reason gives different result, causing issues when we compare Noncartesian to cartesian in some tests
Currently I have added relative tolerance of 1e-3 for it

chaithyagr:
Also, I had a minute issue in gridsearch scripts which came up while using it, the n_coils term is missing and since it is a single line, I am merging it to this PR

zaccharieramzi (approved):

zaccharieramzi:
LGTM @chaithyagr , I'll let you merge when you think it's time

chaithyagr:
merging, I see that the code worked on circleci well.

Issue #112, opened 2020-12-17
[Bug] Normalize k-space locations is not handling `kmax` right.


chaithyagr:
https://github.com/CEA-COSMIC/pysap-mri/blob/d6acae5330c684b66a99d0fbf13e44b0b66046cf/mri/operators/fourier/utils.py#L99-L104

Here, when the the user provides Kmax, we would still want to divide the sample_locations with `2*KMax` to ensure that we are normalizing to `[-0.5, 0.5]` and not `[-1, 1]`

This issue particularly came up when we wanted to send kmax by user for some trajectories. 
I will fix this in a quick PR

PR #113, opened 2020-12-17
Fix normalize_frequency_locations

chaithyagr:
We fix the `normalize_frequency_locations` function as addressed in #112 
This is a quick single line change

Also, in this PR, I am removing PEP8 checks from pytest, as they are redundant, we already have these checks from `pycodestyle`.
Also, it looks like there are some test failures currently due to some change in pytest for PEP8 checks


chaithyagr:
Also, in this PR, I am remove PEP8 checks from pytest, as they are redundant, we already have these checks from `pycodestyle`.
Also, it looks like there are some test failures currently due to some change in pytest for PEP8 checks

zaccharieramzi (approved):
LGTM

PR #114, opened 2021-01-15
Add oscar scale based proximity

chaithyagr:
Hello,

This PR holds
1) Oscar proximity that works scale based
2) Made string compares with `==` rather than `is` which used to cause run errors.
3) Some reformatting of codes.


zaccharieramzi:
Hi @chaithyagr ,

Can you give me more context on what exactly you did regarding the main point of this PR
> Oscar proximity that works scale based
?

I am really not familiar with this so my review may be too shallow. Any pointer would be appreciated.

chaithyagr:
Ahh okay, my bad.
So, currently, the codes only allow for OSCAR regularizer where the prximity is done on every band of wavelet or every coefficient of wavelet or gloablly. (all methods are summarized in a document which I will send off hand)

I have now added scale based proximity. You can ask me off hand for any questions, or we can meet up face to face to speed this process up. For now this PR is not highest priority.

mri/scripts/gridsearch.py
```diff
@@ -262,6 +262,7 @@ def launch_grid(kspace_data, reconstructor_class, reconstructor_kwargs,
                     'samples': fourier_op.samples,
                     'shape': fourier_op.shape,
                     'n_coils': fourier_op.n_coils,
+                    'implementation': 'gpuNUFFT',
```
zaccharieramzi, line 265: can you add a description of this change in the PR description?

mri/operators/proximity/ordered_weighted_l1_norm.py
```diff
@@ -105,36 +129,65 @@ def _op_method(self, data, extra_factor=1.0):
...
+                stop = start + band_size
                 reshaped_data[:, start:stop] = np.reshape(
                     band_data,
-                    (n_channel, np.prod(band_shape_idx)))
+                    (n_channel, band_size)
+                )
+                start = stop
+            output = np.asarray(reshaped_data).T
+            return np.asarray(output).T
+        elif self.mode == 'scale_based':
```
zaccharieramzi, line 159: I feel that there is a lot of mutualisation that could happen between `scale_based` and `band_based`,
If I understand what they are correctly, then it's only normal that this would happen, and I think it should be reflected by the code.

The only difference is more or less the initial reshaping of the data. Even the final one can be largely refactored to be shared.

(Maybe even the `coef_based` could be included but I don't want to get ahead of myself).

zaccharieramzi (changes requested):
a small thing on refactorization ow LGTM (but not a specialist of OSCAR)

mri/scripts/gridsearch.py
chaithyagr: Oh noooo, this was not supposed to be here, i assumed this PR is closed.. This is another issue. I will get another PR for this..

mri/operators/proximity/ordered_weighted_l1_norm.py
chaithyagr: Ahhh, i get your point, ill see what i can do.

chaithyagr: @zaccharieramzi I have made changes to tackl;e this redudancy, but sure made the code a little more complicated, do let me know if it is harder to understand.

zaccharieramzi (approved):
LGTM

PR #115, opened 2021-02-16
B0 correction operator and coefficients

Daval-G:
Implement a B0-correcting pseudo-Fourier operator as a wrapper for any Fourier operator, with in addition different efficient/historical functions to compute the correction coefficients (used by the wrapper).

The pseudo-Fourier wrapper needs a B0 field map (in Hz) and a vector containing all times (in s) after pulse for each trajectory point. Note that all similar shots are handled as one to save computations, but more challenging trajectories can also be handled.

Some future contributions on this topic would be:
1) an automatic option to choose the number of interpolators L, according to empirical results
2) approximated SVD coefficient computations, for the case of demanding trajectories
3) if relevant, an option to visualize the approximation error map

I consider the operator's name temporary, and invite you to propose a more fitting one.


chaithyagr:
@Daval-G do you have the rights to check why the codes are failing? Please do ensure that the codes are passing. I will review this in meantime. Do let me know if this is still Work in progress PR, in which case I would rather just wait.

.travis.yml
```diff
@@ -38,7 +38,7 @@ install:
     - pip install coverage nose pytest pytest-cov coveralls pycodestyle
     - pip install git+https://github.com/CEA-COSMIC/ModOpt.git
     - pip install git+https://github.com/AGrigis/pysphinxdoc.git
-    - pip install sphinx==2.2.1
+    - pip install sphinx==2.2.1 sphinxcontrib-bibtex
```
chaithyagr, line 41: I would prefer to have this on a new line, that way it is clearly visible of what are the things needed to build a doc.

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+
+import numpy as np
+import sklearn.cluster as sc
+
+from .utils import (compute_mfi_coefficients,
+                    compute_mti_coefficients,
+                    compute_svd_coefficients)
+
+
+class ORCFFTWrapper():
```
chaithyagr, line 24: Can we add some tests to check the correctness of this wrapper? Something like the output of this and normal FFT is same in absence of resonance artifacts. Note that you can add cpu based tests (which run on CI) and also gpu based tests (which we plan to run in future in CI, but for now can be run locally for checking correctness)

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+
+    """
+    def __init__(self, fourier_op, field_map, time_vec, mask,
+                 coefficients="svd", weights="full", L=15, n_bins=1000):
+        """ Initialize and compute multi-linear correction coefficients
+
+        Parameters
+        ----------
+        fourier_op: OperatorBase
+            FFT operator to wrap
```
chaithyagr, line 40: I think it will be helpful to mention this can be cartesian or non cartesian

chaithyagr (changes requested):
Some minute concerns. Will continue review soon..

Daval-G:
No need to consider it as a WIP, I will push propositions 1) and 2) soon enough as it's already implemented on my side. I will also add a test for both Cartesian and non-Cartesian wrapping at this occasion.

About the building fails, I am not sure how it relates to this PR. From what I see, one unrelated test is timed out. The same build is valid when run on my side.

chaithyagr:
Also a side note @Daval-G , having tests and a minimum example to know how to use this class will be helpful. I am not sure how much effort it is to get it, we can get only tests here and get another PR for examples.

chaithyagr:
> Also a side note @Daval-G , having tests and a minimum example to know how to use this class will be helpful. I am not sure how much effort it is to get it, we can get only tests here and get another PR for examples.

@Daval-G  Did you plan to get an example in this PR or next one?

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+from .utils import (compute_mfi_coefficients,
+                    compute_mti_coefficients,
+                    compute_svd_coefficients,
+                    compute_fsvd_coefficients)
+
+
+class ORCFFTWrapper(OperatorBase):
+    """ Off-Resonance Correction FFT Wrapper
+
+    This class is used to wrap any FFT operator and change it
```
chaithyagr, line 28:
```suggestion
    This class is used to wrap any Fourier operator and change it
```
I feel FFT gives an idea that it works only with FFTs (cartesian)

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+    into an off-resonance correction multi-linear operator using
+    the method described in :cite: `sutton2003` with different
+    choices of coefficients described in :cite: `fessler2005`.
+
+    """
+    def __init__(self, fourier_op, field_map, time_vec, mask,
+                 coefficients="svd", weights="full", L="auto", n_bins="auto"):
```
chaithyagr, line 35: I feel its better to have something more meaningful than `L`, maybe `num_interpolators` ?

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+        """ Initialize and compute multi-linear correction coefficients
+
+        Parameters
+        ----------
+        fourier_op: OperatorBase
+            Cartesian or non-Cartesian Fourier operator to wrap
+        field_map: numpy.ndarray
+            B0 field inhomogeneity map (in Hz)
```
chaithyagr, line 43: I remember seeing that getting this map up was a painful task, do you think we should have this loader as a utils function here? Or shall we make it a part of `reconstruction_pipelines`?

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+        time_vec: numpy.ndarray
+            1D vector indicating time after pulse in each shot (in s)
```
chaithyagr, line 45: Again, a utils function to get this time vector if we know the start of shot, the TE etc can be helpful. Possibly with defaults set for start time and TE as well.

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+        self.field_map = field_map
+        scale = (range_w[1] - range_w[0]) / self.n_bins
+        scale = scale if (scale != 0) else 1
+        self.indices = np.around((field_map - range_w[0]) / scale).astype(int)
+        self.indices = np.clip(self.indices, 0, self.n_bins - 1)
+
+        # Compute the E=BC factorization and reformat B
+        self.B, self.C, self.E = self.compute_coefficients(field_map, time_vec,
+                                                           mask, L, weights,
+                                                           n_bins)
```
chaithyagr, line 102:
```suggestion
        self.B, self.C, self.E = self.compute_coefficients(
            field_map, 
            time_vec,
            mask, 
            L, 
            weights,
            n_bins,
        )
```

I feel this is cleaner

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # Prepare B to match fourier.op shape
+        if (hasattr(fourier_op, "mask")):
+            self.B = np.tile(self.B,
+                             (fourier_op.mask.size // self.B.shape[0], 1))
+            self.B = self.B.reshape((*(fourier_op.mask.shape), self.L))
+        else:
+            self.B = np.tile(self.B,
+                             (self.samples.shape[0] // self.B.shape[0], 1))
```
chaithyagr, line 111:
```suggestion
            self.B = np.tile(
                self.B,
                (self.samples.shape[0] // self.B.shape[0], 1)
            )
```

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # Compute the E=BC factorization and reformat B
+        self.B, self.C, self.E = self.compute_coefficients(field_map, time_vec,
+                                                           mask, L, weights,
+                                                           n_bins)
+
+        # Prepare B to match fourier.op shape
+        if (hasattr(fourier_op, "mask")):
+            self.B = np.tile(self.B,
+                             (fourier_op.mask.size // self.B.shape[0], 1))
```
chaithyagr, line 107:
```suggestion
            self.B = np.tile(
                self.B,
                (fourier_op.mask.size // self.B.shape[0], 1)
            )
```

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+        x: numpy.ndarray
+            input N-D array with the same shape as fourier_op.shape
+        Returns
+        -------
+            masked distorded Fourier transform of the input volume
+        """
+        y = 0
+        for l in range(self.L):
+            y += self.B[..., l] * self.fourier_op.op(
+                                    self.C[l, self.indices] * x, *args)
```
chaithyagr, line 134: Please try to follow the same code style as above..
```python
fun(
    lll,
    qqwqw,
    sasasa,
)
```

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+        ----------
+        x: numpy.ndarray
+            masked distorded N-D k-space
+        Returns
+        -------
+            inverse Fourier transform of the distorded input k-space.
+        """
+        y = 0
+        for l in range(self.L):
+            y += np.conj(self.C[l, self.indices]) * self.fourier_op.adj_op(
```
chaithyagr, line 152: Here too

mri/operators/fourier/utils/orc_coefficients.py
```diff
@@ -0,0 +1,278 @@
...
+    -------
+    centers: numpy.ndarray
+        New bin centers of the reduced histogram
+    """
+
+    # Compute kmeans to get custom centers
+    km = sc.KMeans(n_clusters=L, random_state=0)
+    km = km.fit(samples.reshape((-1, 1)), sample_weight=counts)
+    centers = np.array(sorted(km.cluster_centers_)).flatten()
+
```
chaithyagr, line 83: Why the extra space?
```suggestion
```

mri/operators/fourier/utils/orc_coefficients.py
```diff
@@ -0,0 +1,278 @@
...
+def compute_mfi_coefficients(field_map, time_vec, mask, L=15,
+                             weights="full", n_bins=1000):
+    """
+    This function implements the correction weights described
+    in :cite: `man1997` and extended in :cite: `fessler2005`
+    called 'Multi-frequency interpolation' coefficients
+
+    Parameters
+    ----------
+    field_map: numpy.ndarray
```
chaithyagr, line 96: I dont think we need to repeat the documentation. Let us just have these as internal function and ask the user to refer to the ORC Wrapper for details. For now, I feel this just adds maintainence overhead. Its your call..
We can later think of how to repeat the documentation in sphinx generation pipelines. I am sure that it is doable.

chaithyagr: YOu can change function name to `def _compute_...` in case you make this internal function and you are sure that no user will use it.

mri/tests/test_orc_wrapper.py
```diff
@@ -0,0 +1,250 @@
...
+            Nz = sampling_z.sum()
+            z_locations = np.repeat(
+                convert_mask_to_locations(sampling_z),
+                samples.shape[0],
+            )
+            z_locations = z_locations[:, np.newaxis]
+            kspace_loc = np.hstack([np.tile(samples, (Nz, 1)), z_locations])
+            print(kspace_loc.shape)
+            print("Process Stacked3D-FFT test in 3D '{0}'...", i)
+            """
```
chaithyagr, line 200: Why is this commented?

chaithyagr (changes requested):
Some review comments, renames and minute codestyle changes. Please try to have this update codestyle wherever possible

chaithyagr:
Hey @Daval-G , can we have this up by december 10th? Things to do in my opinion:

- [ ] Rebase
- [ ] FIx any codestyle issues
- [ ] Testing and submit

mri/operators/fourier/orc_wrapper.py
Daval-G: It depends deeply on how the field map is stored, so I don't think it should be part of pysap-mri.

mri/operators/fourier/orc_wrapper.py
Daval-G: Similarly to the field map, it deeply depends on the trajectory and might not be as trivial as our usual case.

mri/operators/fourier/orc_wrapper.py
```diff
@@ -0,0 +1,150 @@
...
+            1D vector indicating time after pulse in each shot (in s)
+        mask: numpy.ndarray
+            Mask describing the regions to consider during correction
+        coefficients: {'svd', 'mti', 'mfi'}, optional
+            Type of interpolation coefficients to use (default is 'svd')
+        weights: {'full', 'sqrt', 'log', 'ones'}
+            Weightning policy for the field map histogram (default is "full")
+        num_interpolators: int, "auto"
+            Number of interpolators used for multi-linear correction
+        n_bins: int, "auto"
```
chaithyagr, line 52: I think we need to explain what "auto" means.. For both cases

chaithyagr (approved):
MInute point, else LGTM. Just adress this minute point and we can merge when green

chaithyagr (approved):
i think we can merge.

Issue #117, opened 2021-03-09
CircleCI tests are running for too long


chaithyagr:
Currently, many PRs are blocked particularly because the examples are running for too long.

We need to work with this as many times the things just time out and fail

chaithyagr:
I see that the major culprit for this is the example `non_cartesian_reconstruction_stack_3d.py`, which was introduced in the days when we didnt have gpuNUFFT and we carried out reconstructions of 3D NUFFT for stacks by 2D plane NUFFT and 1D FFT.
While we still feel this NUFFT could be helpful and will hold on to it for a bit longer and test it with tests (Fourier Adjoint test with stackedFourier), I feel we can remove this example as it misguides people into thinking that this is the best case scenario. 

What do you think @zaccharieramzi ?

chaithyagr:
Further, I am adding here that, this code caused issues even in past, so in examples we limited it to 10 iterations, due to which the final result is worse than a simple gridded solution.

zaccharieramzi:
How much longer is it compared to the rest of the tests?

I see that in circle ci we are not using a matrix to divide up the testing, so maybe doing that first would reduce the time taken.

chaithyagr:
Well for now it takes more than 10 minutes to just get the lipchitz constant. This is timeout after which circleci closes the testing.
Yes surely adding matrix could be helpful here.

chaithyagr:
Locally it isnt very long, but that is cause we carry out fourier along z (totally NxN 1D FFTs), and we can afford a lot of threads locally, while circleci fails with it..

zaccharieramzi:
I see. Then indeed let's not run this example.
But I suggest also adding in the matrix.

chaithyagr:
Yes, will add it on to #118

PR #116, opened 2021-02-22
Fix gridsearch script for non cartesian

chaithyagr:
The Non Cartesian operator needs to be created with kwargs as they cant be pickled while running with joblib. 

Earlier we passed only some arguments, now we added to track for all the arguments.

PR #118, opened 2021-03-11
Remove examples and remove unwanted imports

chaithyagr:
This PR handles #117 
We just remove a long example so that circleci doesnt block PRs.
Also remove some unwanted imports which seemed to be present in `cartesian.py` and which surely will confuse the users.
Finally, we also ensure that we run 4 jobs in parallel on circleci.


examples/non_cartesian_reconstruction_stack_3d.py
```diff
@@ -1,122 +0,0 @@
-"""
```
zaccharieramzi, line 1: I think we should keep this example and only run it if certain conditions are met. For example it could be:
- sufficient number of cpu cores
- not being in CI

I guess all of these conditions could be checked via environment variables

.circleci/config.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     docker:
       - image: circleci/python:3.6
     working_directory: ~/repo
+    parallelism: 4
```
zaccharieramzi, line 12: why 4?

.circleci/config.yml
```diff
@@ -43,7 +44,9 @@ jobs:
             pip install --upgrade pip
             pip install Cython
             pip install -e .
-            python .circleci/test_script.py
+            TESTFILES=$(circleci tests glob examples/*.py | circleci tests split --split-by=timings)
+            echo ${TESTFILES}
+            python3 -m unittest ${TESTFILES}
```
zaccharieramzi, line 49: Why unittest?

.circleci/config.yml
```diff
@@ -43,7 +44,9 @@ jobs:
             pip install --upgrade pip
             pip install Cython
             pip install -e .
-            python .circleci/test_script.py
+            TESTFILES=$(circleci tests glob examples/*.py | circleci tests split --split-by=timings)
```
zaccharieramzi, line 47: Is this the thing allowing to do a matrix in circle ci?

zaccharieramzi (changes requested):
I think we should keep the example and just not run it in CI

examples/non_cartesian_reconstruction_stack_3d.py
chaithyagr: yes we can do that I think

.circleci/config.yml
chaithyagr: No particular reason, let me see if I can scale up more

.circleci/config.yml
chaithyagr: Again, no particular reason. I just wanted a way to run all the python scripts at once..

.circleci/config.yml
chaithyagr: This is what will split the python scripts into 4 batches, based on previous run times..

zaccharieramzi: oh nice that it's based on previous run times ! Amazing

.circleci/config.yml
zaccharieramzi: ok sure

.circleci/config.yml
chaithyagr: There doesnt seem to be a maximum option... Why not just keep it 4 with scope for improving?

zaccharieramzi: no but typically we could have it at 7, the number of examples to be run.

chaithyagr: oh that may not be the best thing, cause every container builds all the environments to just run one test. Thats too much overhead and maybe too much unwanted resource usuage. Right now duration is nearly 23min, maybe I will see how much gain we can expect and then decide..

zaccharieramzi: do we have a hard-limit on resources, or is this a purely reasonable concern?

chaithyagr: Well i think we have a limit of 2,500 free credits/week on circle ci:  https://circleci.com/pricing/#
The way how many credits is used by us, I have no idea really..

zaccharieramzi: ok let's not spend too much time on that we can stick with 4, I am fine with it. We should consider switching to GH actions which is clearer and more integrated.

chaithyagr: I agree

examples/non_cartesian_reconstruction_stack_3d.py
chaithyagr: Back at this, I feel this will take longer, can we merge for now and get at this later? esp as this blocks #116 #114 ?

zaccharieramzi: Ok, let's then open an issue to reintegrate this, where we can copy-paste the content of this file.

zaccharieramzi (approved):
LGTM, you can merge when green

PR #121, opened 2021-05-05
Fix the #66 issue. This will ensure that we dont enforce sparsity in approximation

### LINKED ISSUE #66: Better sparsity promotion
Right now sparsity is promoted by an L1 Norm on all the wavelet coefficients in the examoles. However the coarse scale of the wavelet transform has no reason to be sparse and shouldn't be biased.

The examples should highlight that. For this, it might be necessary to have a helper function which for a given `mu` and a given wavelet transform returns the corresponding thresholding levels for all coefficients (the ones corresponding to the coarse scale being 0)

chaithyagr:
Resolves #66 

This is a minor hacky fix, but it should get the job done for now!

While we work on removing flattening, I feel this is a good first step to ensure we don't end enforcing sparsity on coarse-scale!

Another option to set up a regularizer with 0 weights for approximation, but that still keeps computation hard! Not to mention it also has issues that it depends on user to do it.

### DIFF
--- a/mri/operators/linear/wavelet.py
+++ b/mri/operators/linear/wavelet.py
@@ -55,6 +55,7 @@ class WaveletN(OperatorBase):
         self.unflatten = unflatten
         self.n_jobs = n_jobs
         self.n_coils = n_coils
+        self.coarse = None
         if self.n_coils == 1 and self.n_jobs != 1:
             print("Making n_jobs = 1 for WaveletN as n_coils = 1")
             self.n_jobs = 1
@@ -92,10 +93,10 @@ class WaveletN(OperatorBase):
         transform = self.transform_queue.pop()
         transform.data = data
         transform.analysis()
-        coeffs, coeffs_shape = flatten(transform.analysis_data)
+        coeffs, coeffs_shape = flatten(transform.analysis_data[1:])
         # Add back the transform to the queue
         self.transform_queue.append(transform)
-        return coeffs, coeffs_shape
+        return coeffs, coeffs_shape, transform.analysis_data[0]
 
     def op(self, data):
         """ Define the wavelet operator.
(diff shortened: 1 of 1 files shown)


mri/operators/linear/wavelet.py
```diff
@@ -125,10 +126,10 @@ def op(self, data):
             )
             coeffs = np.asarray(coeffs)
         else:
-            coeffs, self.coeffs_shape = self._op(data)
+            coeffs, self.coeffs_shape, self.coarse = self._op(data)
         return coeffs
 
-    def _adj_op(self, coeffs, coeffs_shape, dtype="array"):
+    def _adj_op(self, coeffs, coeffs_shape, coarse, dtype="array"):
```
zaccharieramzi, line 132: don't forget to add it to the docs

mri/operators/linear/wavelet.py
```diff
@@ -125,10 +126,10 @@ def op(self, data):
             )
             coeffs = np.asarray(coeffs)
         else:
-            coeffs, self.coeffs_shape = self._op(data)
+            coeffs, self.coeffs_shape, self.coarse = self._op(data)
```
zaccharieramzi, line 129: Are there not parts of the code which depend on getting all the coefficients?

zaccharieramzi (changes requested):
One q and one remark

mri/operators/linear/wavelet.py
philouc: Absolutely zaccharie you're right!

mri/operators/linear/wavelet.py
chaithyagr: Hmmm not from what I recall. Let me add another method to get the coefficients anyway.
(esp currently we only return flattened versions)

chaithyagr:
Looks like this wont work after all. In optimization we subtract the values of detail, however as we retain the same approximation coefficients, we in the end just mess up the results!
The only way forward is to introduce the weights such that we dont modify the coarse scale!

I am getting that PR up. Closing this here...

PR #123, opened 2021-06-09
Add Weighted Sparse Threshold

### LINKED ISSUE #66: Better sparsity promotion

chaithyagr:
This resolves #66 

Weighted Sparse threshold will ensure that we enforce sparsity only on non-coarse scales.

It also allows for users to give scale dependent regularization and custom regularizations.


chaithyagr:
Tagging for mid 2021 release

mri/operators/proximity/weighted.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+
+from modopt.opt.proximity import SparseThreshold
+
+
+class WeightedSparseThreshold(SparseThreshold):
+    """This is a weighted version of `SparseThreshold` in ModOpt.
+    When chosen `scale_based`, this applied weighted proximity as :
+        W * P^i where i is the scale
```
zaccharieramzi, line 9: Not clear here what `W` and `P^i` are

zaccharieramzi: I think I know understand given what's underneath, but I don't understand where this rule comes from, do you have a reference?

zaccharieramzi (changes requested):
It would be also nice to have unit tests for the newly introduced prox op. Right now we only test a very simple setting

mri/operators/proximity/weighted.py
chaithyagr: Ill try to add it.

chaithyagr:
For test, i really didnt know how to get it up.. Maybe just check the underlying weights obtained finally?

mri/operators/proximity/weighted.py
chaithyagr: @philouc can you please help me find a way to get the reference for variable scale? I just kept it as generic as possible, but I know for a fact that 1  / \lamda ^ s where s is scale is used constantly..
Here we can set any base scale and any power (as a geometric progression)

chaithyagr:
I have repurposed the scale based to have custom weights per scale to prevent confusion and make it more generic. The power law weights can be set using `np.geomspace` directly while passing input so i dont think it will be that much of a problem.

mri/operators/proximity/weighted.py
philouc: Hey Chaithya, I have to find it in Pesquet & Chouzenoux papers as in Loubna 2019 ISBI's we did not cite the exact reference even though we wrote the right formulae in Eq. (4)

mri/tests/test_proximity.py
```diff
@@ -0,0 +1,66 @@
...
+import numpy as np
+
+# Package import
+from mri.operators import WaveletN
+from mri.operators.proximity.weighted import WeightedSparseThreshold
+
+
+class TestProximity(unittest.TestCase):
+    # Test proximity operators
+    def test_weighted_sparse_threshold(self):
```
zaccharieramzi, line 21:
```suggestion
    def test_weighted_sparse_threshold_weights(self):
```

mri/tests/test_proximity.py
```diff
@@ -0,0 +1,66 @@
...
+import unittest
+import numpy as np
+
+# Package import
+from mri.operators import WaveletN
+from mri.operators.proximity.weighted import WeightedSparseThreshold
+
+
+class TestProximity(unittest.TestCase):
+    # Test proximity operators
```
zaccharieramzi, line 20: I think it could be nice to add at least one thing, verify that the application of `op` runs smoothly.

The cherry on top would be to verify that the results make sense in a very small scale example.

zaccharieramzi (commented):
Some remarks on testing

mri/tests/test_proximity.py
chaithyagr: Haha, will see what i can do.. Although, in my opinion that is checked in modopt...

chaithyagr:
The more i think about this, I feel this should become a modification for Sparse Threshold itself! Why have a weighted version doing pretty much same stuff as Sparse threshold separately? 
@sfarrens what do you think? (To give you context, we needed a version of SparseThreshold where we can send weights based on scales or custom weights. SparseThreshold only supports the latter right now if I am right.)

zaccharieramzi:
I think because this is rather wavelets oriented, so it shouldn't live in ModOpt. However it could definitely have its place in pysap rather than pysap-mri. Wdyt?

chaithyagr:
If thats the case, we should even move the WaveletN class to `pysap`. I wanted to do this earlier in fact, but it needed more people signing off, and hence I didnt do it. How about we add this class here for now, so that we have a working solution. In the next release, we can do the movement of the common classes to `pysap`, after prior discussions.

chaithyagr:
I feel for the best case scenario, lets have this in pysap-mri for now. The major refactoring out codes from plugins can be done in the next release. @zaccharieramzi WDYT?

zaccharieramzi:
Sure @chaithyagr , if it's easier why not. I think my outsanding issues still remain unanswered though.

chaithyagr:
Yes getting them fixed now

mri/operators/proximity/weighted.py
```diff
@@ -0,0 +1,77 @@
...
+    weights : numpy.ndarray
+        Input array of weights or a tuple holding base weight W and power P
+    coeffs_shape: tuple
+        The shape of linear coefficients
+    weight_type : string 'custom' | 'scale_based' | 'custom_scale',
+        default 'scale_based'
+        Mode of operation of proximity:
+        custom       -> custom array of weights
+        scale_based -> custom weights applied per scale
+    zero_weight_coarse: bool, default True
```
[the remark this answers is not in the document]
chaithyagr: Done

mri/tests/test_proximity.py
chaithyagr: Added very minute test. As I stated above the other tests will be done in Modopt as this is just a wrapper on Sparse Threshold.

zaccharieramzi (approved):
LGTM

PR #128, opened 2021-10-01
Change kspace locations normalization to properly wrap values inside the 0.5-circle.

Daval-G:
Change the `normalize_frequency_locations` function to tackle two issues relatively to the behavior when normalized locations are outside the ]-0.5, 0.5[ interval:
(1) Only positive values were checked, while both positive and negative values would cause problems.
(2) Only the locations strictly at the limit of the circle were wrapped, and not the ones outside.


chaithyagr (approved):
LGTM, merge when green.

PR #131, opened 2021-11-30
remove pyNUFFT

paquiteau:
This PR remove pyNUFFT. Its internal API have changed and gpuNUFFT is better anyway. 

"There should be one-- and preferably only one --obvious way to do it."


mri/operators/fourier/non_cartesian.py
```diff
@@ -567,11 +331,6 @@ def __init__(self, samples, shape, implementation='cpu', n_coils=1,
             self.density_comp = density_comp
```
chaithyagr, line 331: Kindly change the documentation also 
`implementation: str 'cpu' | 'cuda' | 'opencl' | 'gpuNUFFT',` We need to remove 'opencl' and 'cuda'

chaithyagr (changes requested):

chaithyagr (approved):
LGTM

PR #133, opened 2021-12-01
Generators

paquiteau:
Splittings things up:
This PR add support for Generators of Kspace (iterate over the available acquisition data). These can be use with ColumnFFT Operator to reduce memory footprint and speed up computations.


mri/generators/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .base import KspaceGeneratorBase
```
chaithyagr, line 1: Do we have to import the Base class?

mri/generators/base.py
```diff
@@ -0,0 +1,94 @@
+"""
+Base Class for KspaceGenerators.
```
chaithyagr, line 2:
```suggestion
Base Class for k-space generators.
```

Note that KspaceGenerators is the base class right?

mri/generators/base.py
```diff
@@ -0,0 +1,94 @@
...
+        if self.iter < self._len:
+            self.iter += 1
+            return self._full_kspace, self.mask
+        raise StopIteration
+
+    def reset(self):
+        """Reset the Generator to its initial state."""
+        self.iter = 0
+
+    def opt_iterate(self, opt, reset=True, estimate_call_period=None):
```
chaithyagr, line 78: Please document this.

mri/generators/base.py
```diff
@@ -0,0 +1,94 @@
...
+            opt.idx += 1
+            opt._grad.obs_data = kspace
+            opt._grad.fourier_op.mask = col
+            opt._update()
+            if opt.metrics and opt.metric_call_period is not None:
+                if opt.idx % opt.metric_call_period == 0 or opt.idx == (self._len - 1):
+                    opt._compute_metrics()
+            if estimate_call_period is not None:
+                if opt.idx % estimate_call_period == 0 or opt.idx == (self._len - 1):
+                    x_new_list.append(opt.get_notify_observers_kwargs()["x_new"])
```
chaithyagr, line 92: Isnt this similar to what is done in modopt? Isnt it redundant?

mri/generators/cartesian.py
```diff
@@ -0,0 +1,94 @@
...
+from .base import KspaceGeneratorBase
+
+
+class Column2DKspaceGenerator(KspaceGeneratorBase):
+    """k-space Generator, at each step a new column fills the existing k-space."""
+
+    def __init__(self, full_kspace, mask_cols, max_iter=0):
+        mask = np.zeros(full_kspace.shape[-2:])
+
+        def flip2center(mask_cols, center_pos):
```
chaithyagr, line 19: Maybe a better name / documentation? Something that reflects what actually is being done in MRI context. If I am right you are basically sampling from center of k-space and then moving outward slowly (while you flip)

mri/generators/cartesian.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __next__(self):
+        if self.iter > self._len:
+            raise StopIteration
+        idx = min(self.iter + 1, len(self.cols))
+        self.iter += 1
+        return self._get_kspace_and_mask(idx)
+
+
+class OneColumn2DKspaceGenerator(Column2DKspaceGenerator):
```
chaithyagr, line 59: Theres a lot of commonalities in this and above class, can we have a single class with a variable send in init if we want all data or just next data? 

To me personally, Gadgetron works based on `OneColumn2DKspaceGenerator` and only this makes sense as a use-case.  `Column2DKspaceGenerator` must only be a wrapper in case the algorithm needs it but not a separate generator. WDYT?

mri/generators/cartesian.py
```diff
@@ -0,0 +1,94 @@
...
+        return self._get_kspace_and_mask(idx)
+
+    def _get_kspace_and_mask(self, idx: int):
+        mask = np.zeros(self.shape[-2:])
+        mask[:, self.cols[idx]] = 1
+        kspace = np.squeeze(self._full_kspace * mask[np.newaxis, ...])
+        return kspace, mask
+
+
+class DataOnlyKspaceGenerator(OneColumn2DKspaceGenerator):
```
chaithyagr, line 86: Again, can be same class, extra arg. Lets not make a lot of classes as then we will have to document a lot and also makes codebase quite big.

chaithyagr: Dont forget to document the new arg also

mri/tests/test_generators.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_kspace_generator(self):
+        mask_cols = np.random.randint(0, 64, size=16)
+        for nc in self.n_coils:
+            full_kspace = np.squeeze(np.random.rand(nc, *self.shape))
+            data_gen = DataOnlyKspaceGenerator(full_kspace, mask_cols)
+            data_gen2 = OneColumn2DKspaceGenerator(full_kspace, mask_cols)
+            data_gen3 = Column2DKspaceGenerator(full_kspace, mask_cols)
+
+            for (line, col), (kspace, mask), (kspace2, mask2) in zip(data_gen, data_gen2, data_gen3):
```
chaithyagr, line 34: Maybe test against ground truth (i.e. from `full_kspace`) rather than `DataOnlyKspaceGenerator` ?

Also, do add a check for the mask as well.

mri/tests/test_generators.py
```diff
@@ -0,0 +1,53 @@
...
+        mask_cols = np.random.randint(0, 64, size=16)
+        for nc in self.n_coils:
+            full_kspace = np.squeeze(np.random.rand(nc, *self.shape))
+            data_gen = DataOnlyKspaceGenerator(full_kspace, mask_cols)
+            data_gen2 = OneColumn2DKspaceGenerator(full_kspace, mask_cols)
+            data_gen3 = Column2DKspaceGenerator(full_kspace, mask_cols)
+
+            for idx, (line, col) in enumerate(data_gen):
+                print(idx, col)
+                np.testing.assert_equal(data_gen[idx][0], data_gen3[idx+1][0][..., col])
```
chaithyagr, line 50: Same as above

chaithyagr (changes requested):
SOme minor concerns in class writing, and minute documentation and testing

mri/generators/__init__.py
paquiteau: We could do it without it, but there is a lot in common to generators. Currently only cartesian one has been develop, but extending to non-cartesian by inheriting the Base class will be  easier.

mri/generators/base.py
paquiteau: Forward-Backward and primal-dual algorithms in modopt indeed have  an `iterate` methods. 
Here the idea is that it is the generator, which drive the reconstruction, as it is this class that yields the data required for it. This code could also be  in `run_online_algorithm` (see PR #134), but I find it more relevant here. One could for instance easily modify this method to provide an another way of iterating on the data (for instance doing a pre-processing step before calling `opt._update`).

mri/generators/cartesian.py
paquiteau: You are right.

mri/generators/cartesian.py
paquiteau: I agree. I am going to do the relevant changes.

mri/generators/cartesian.py
paquiteau: I disagree. Here the shape of the data is different than the one from `OneColumn2DKspaceGenerator` or `Column2DKspaceGenerator`. This approach is different, being memory efficient and thus incompatible with the offline fourier operators. So I prefer to have it separated.

chaithyagr: No, but from what I see, all of them are under the purview of the same functionality. The current implementation is like having a separate class for each kind of NUFFT (while in reality with have just one NonCartesianFFT with `implementation` parameter.

mri/generators/base.py
chaithyagr: How about something like:
```python
opt._grad.obs_data = kspace
opt._grad.fourier_op.mask = col
opt.iterate(num_iterations=1) # 1 or any number of iterations, to me it must be an argument that can be set. 
                                               # In reality we want as many iterations as need till we get next k-space data
```

That will reduce a lot of the redundant codes.. 

```

File: mri/tests/test_generators.py (hunk 9 above)
Comment (paquiteau) [reply mri/tests/test_generators.py]: Done.

File: mri/tests/test_generators.py (hunk 8 above)
Comment (paquiteau) [reply mri/tests/test_generators.py]: Done.

File: mri/generators/cartesian.py (hunk 7 above)
Comment (paquiteau) [reply mri/generators/cartesian.py]: Yep, Had the same though, you were right, it's done

File: mri/generators/cartesian.py (hunk 6 above)
Comment (paquiteau) [reply mri/generators/cartesian.py]: done.

File: mri/generators/cartesian.py [hunk 10]
```diff
@@ -0,0 +1,97 @@
...
+    mode: {"line", "current", "memory"}
+        If "line" : at step k, the generator yields only the
+        data of k-th column, and the column number associated.
+        If "current": at step k, the generator yields an array
+        with the shape of full_kspace, but containing only the data
+        of step k, and the mask associated.
+        If "memory", same as "current", but all the data that have
+        been previously acquired fills the kspace-array.
+        Default is "current".
+    start_center: bool
```
chaithyagr, line 34:
```suggestion
    start_center: bool, default `True`
```

mri/operators/fourier/online.py
```diff
@@ -0,0 +1,128 @@
...
+            val_shift = (self.shape[1] // 2 + val) % self.shape[1]
+        if val >= self.shape[1]:
+            raise IndexError("Index out of range")
+        self._mask = val
+        cos = np.cos(2 * np.pi * val_shift / self.shape[1])
+        sin = np.sin(2 * np.pi * val_shift / self.shape[1])
+        exp_f = cos - 1j * sin
+        exp_b = cos + 1j * sin
+        self._exp_f = (1 / np.sqrt(self.shape[1])) * (exp_f ** np.arange(self.shape[1]))
+        self._exp_b = (1 / np.sqrt(self.shape[1])) * (exp_b ** np.arange(self.shape[1]))
```
chaithyagr, line 81: Maybe document over here on what is being done? I understand its 1D FFT done later, but maybe add what exp_f and exp_b is?

chaithyagr (changes requested):
Minute nitpicks. Else seems good. Thank you for the PR

chaithyagr:
@paquiteau assume I approved the PR after the above 2 requests are tackled.

PR #136, opened 2021-12-01
Add function to discard kspace locations out-of-bounds, and related tests

Daval-G:
As opposed to previous PR #128, kspace samples located outside the [-0.5; 0.5[ limits should simply be discarded. Since this manipulation is not expected from `normalize_kspace_locations` and would represent an unwanted and redundant intermediate step inside the Fourier operators, it was preferred to simply add an utilitary function for the users.

Additional tests were added in `mri/tests/test_fourier_adjoint.py` for 2D and 3D by mimicking the `normalize_frequency_locations` tests. However, the latters had issues in their assertions and all four tests were properly updated to match the intended checks.

### DIFF
--- a/mri/operators/fourier/utils.py
+++ b/mri/operators/fourier/utils.py
@@ -79,8 +79,8 @@ def convert_locations_to_mask(samples_locations, img_shape):
 
 def normalize_frequency_locations(samples, Kmax=None):
     """
-    This function normalize the samples locations between [-0.5; 0.5[ for
-    the non-cartesian case
+    This function normalizes the sample locations between [-0.5; 0.5[ for
+    the non-Cartesian case.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


mri/operators/fourier/utils.py
```diff
@@ -108,6 +108,33 @@ def normalize_frequency_locations(samples, Kmax=None):
...
+    reduced_kspace_loc: np.ndarray
+        The sample locations reduced strictly to [-0.5; 0.5[ by discarding
+        outliers.
+    reduced_kspace_data: np.ndarray
+        The samples corresponding to reduced_kspace_loc defined above.
+    """
+    kspace_mask = np.all((kspace_loc < 0.5) & (kspace_loc >= -0.5), axis=-1)
+    kspace_loc = kspace_loc[kspace_mask]
+    kspace_data = kspace_data[:, kspace_mask]
+    return kspace_loc, kspace_data
```
chaithyagr, line 135: Maybe we need to use np.ascontiguous array for gpuNUFFT here?

chaithyagr (changes requested):
Minute detail

mri/operators/fourier/utils.py
chaithyagr:
```suggestion   
    return np.ascontiguousarray(kspace_loc), np.ascontiguousarray(kspace_data)
```

chaithyagr (changes requested):

chaithyagr:
I just submitted my change @Daval-G , LGTM, merge when green

chaithyagr (approved):

PR #140, opened 2022-01-27
Add support for pygpuNUFFT

paquiteau:
This PR adds bindings of pygpuNUFFT to pysap-mri. It provides handles to gpu-only functions.


paquiteau:
The Failing test are related to wavelet operation precision. I don't know what changed, but its not on my code.

paquiteau:
Let's use https://github.com/paquiteau/mri-cufinufft instead.

PR #157, opened 2022-06-24
Setup the workflow needed for getting API docs

chaithyagr:
This PR will make sure that we will have runnable API docs which can be triggered.

This must be targeted to `master` so that its effect can take place in every other PR (if I am right).

The example of the output is present in actions in my fork: https://github.com/chaithyagr/pysap-mri/actions/runs/2555701992


.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,51 @@
+name: Build API Doc
+
+on: [workflow_dispatch]
+
+jobs:
+  api:
+    name: Deploy API Documentation
+    runs-on: ubuntu-latest
+    env:
+      root: https://github.com/sfarrens/
```
paquiteau, line 10: This should rather be targeting to CEA-COSMIC I guess

.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,51 @@
...
+      tag: /archive/refs/tags/
+      vpysap: 999
+      folder: pysap-999
```
paquiteau, line 13: why 999? Is it a dummy version number ?

paquiteau (commented):
Thanks  for the CI and CD voodoo incantations. I might not be the best to decypher it, but found some minor changes to do.

.github/workflows/test-docs.yml
chaithyagr: Ahhh now maybe yes, initially, it was not yet up

.github/workflows/test-docs.yml
chaithyagr: for now yes.

PR #158, opened 2022-06-30
 Final Documentation Pull Request #155 

chaithyagr:
This the final documentation improvement PR.
All documentation improvement should start from the `documentation` branch, and target it in their PR.
When everything will be correct for the documentation, we will merge this into `develop`.


paquiteau (approved):
LGTM, but the test are failing. from What I have seen there is no change in the active code in this PR, so the problem might goes deeper than that.

PR #166, opened 2024-01-08
Mrinufft

chaithyagr:
Make pysap-mri compatible with mri-nufft


mri/operators/fourier/utils/processing.py
```diff
@@ -347,24 +273,16 @@ def estimate_density_compensation(kspace_loc, volume_shape, num_iterations=10):
...
-                         "estimate the density compensation")
-    grid_op = NonCartesianFFT(
-        samples=kspace_loc,
-        shape=volume_shape,
-        implementation='gpuNUFFT',
-        osf=1,
+    density_comp = get_density(implementation)(
+        kspace_loc,
+        volume_shape,
+        **kwargs
```
paquiteau, line 286: This needs a backend argument as well (to select gpunufft). Otherwise you can also do 
`get_operator("gpunufft").pipe(kspace_loc, volume_shape)`

chaithyagr: Yes, that goes in kwargs. Cause I think at the moment for `cell_count` we dont have backend argument. So I didnt want to fixate on anything specifically

chaithyagr:
I have refactored all the codes and now it is ready for review @paquiteau

examples/gridsearch_launch.py
```diff
@@ -12,7 +12,7 @@
 Wavelets    ==> sym8 and sym12
 nb_scale    ==> 3 and 4
 """
```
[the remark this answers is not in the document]
chaithyagr: I wanted to add a new example for that.

chaithyagr:
I have now rebased to `develop` branch to include auto threshold stuff

chaithyagr:
> Great job ! Some remarks:
> 
> * I think the density examples could be merge into one (to ease the comparison between them).
> * You removed the stacked nufft operator, without providing a replacement (there is one in mri-nufft however), maybe we want a dedicated example for that too.


At the moment I dont see a need for it sadly... We can add one soon when we get to it I guess?

PR #176, opened 2024-03-12
update to src layout and cleanup dependencies.

paquiteau:
TODO: 

- [ ] Apply ruff and black 
- [ ] Cleanup the CI


.github/workflows/ci-build.yml
```diff
@@ -35,7 +35,6 @@ jobs:
         run: |
           conda info
           conda list
-          sudo apt install -y libnfft3-dev
```
chaithyagr, line 38: Why did you remove this? We do need it for testing right?

.github/workflows/ci-build.yml
```diff
@@ -48,7 +47,6 @@ jobs:
           python -m pip install git+https://github.com/CEA-COSMIC/pysap.git@develop
           python -m pip install git+https://github.com/AGrigis/pysphinxdoc.git
           python -m pip install coverage nose pytest pytest-cov pycodestyle pydocstyle twine pytest-xdist
-          python -m pip install pynfft2
```
chaithyagr, line 51: Same as above

chaithyagr (changes requested):
Thank you so much for this PR. I will clean up the pipeline and ensure the tests are not redundant (assuming mri-nufft does most of NUFFT tests). 
Quick questions though:
1) Do we move to finufft as default installation along with pysap-mri, and as the default backend?

paquiteau:
yes finufft should be the default backend i think, its the easiest to install, and has the best performances on cpu.

chaithyagr:
Shall we merge this and the do a PR for pipeline fixes? I think that would be better to avoid too many changes in a single PR

